PR #4, opened 2023-07-11
Version/0.2

ss2165:


acl-cqc:
This looks quite big, and I think I've seen some of this go into main quite recently. Is there some merge/rebase issue?

acl-cqc (approved):
I don't see any reason why any of this can't be pushed, so approved.

But there may be reasons why we might not want to! Of course there are small-scale things that I might have commented on in reviews if I'd looked at the originals, but besides that,
1. There's code in `tests/` that doesn't execute; "python -m pytest ." certainly does stuff, but I can insert things that should break, and all the tests pass. (E.g. I tried `sys.exit(-1)` inside `test_worker`'s `id_with_callback` and everything was happy. However, the same inside `id_delay` *did* break tests, so I can see that separating out the code that does test the python from the rest could be tricky)
2. The big thing I find myself wanting is documentation as to what the python runtime supports and what it doesn't. It's one thing when there are two runtimes (with different feature sets), and the proto's define an interface for the union of both (which is actually equal to the feature-set of one runtime); it's another when there's only one runtime, and the proto's define an interface with a bunch of extra fields that have no explanation!

PR #5, opened 2023-09-20
changes from QPL demo

ss2165:


tests/conftest.py
```diff
@@ -53,6 +60,12 @@ def pytest_collection_modifyitems(config, items):
 
 
 def pytest_configure(config):
+    config.addinivalue_line(
+        "markers",
+        "skip_typecheck: this mark skips tests that"
+        " require typecheck installed if it is not installed.",
```
acl-cqc, line 66: The only 'nit' I can see is I'd rephrase this to avoid the double "installed" - "require the typecheck module, if it is not installed", say

acl-cqc (approved):
All looks fine to me....I guess we never put in a negative constraint on pytket version?

PR #9, opened 2024-02-02
Release/0.4.0

ss2165:


acl-cqc (approved):
Yup ok, looks good. I'm not clear that even [Buf](https://buf.build/) would allow us to move `jobs.proto` out of here and the other hacks would be far worse....

PR #10, opened 2024-02-15
Release/0.5.0

ss2165:


acl-cqc (approved):
Sounds like a good point to make a release. If we can cut down on any of the generics-handling code then that might be a minor/backwards-compatible release...

PR #14, opened 2024-04-10
Release/0.7.0

ss2165:
Includes commits from releases 0.6, 0.6.1, 0.7.0

I suggest we close #12 and once this is merged I can go through and tag/release the right commits for all three versions.


tests/test_type_check.py
```diff
@@ -258,7 +259,7 @@ def ifelse() -> Output:
     @lazy_graph(type_check_sig=sig)
     def lop() -> Output:
         @loop()
-        def lopdef(x) -> Output:
+        def lopdef(x):
```
acl-cqc, line 262: Heh, missed a few here, these could have stayed, ah well

tests/test_builder.py
```diff
@@ -181,31 +164,31 @@ def g() -> Output[IntValue]:
 @pytest.mark.parametrize(
     "builder,expected_gen",
     [
-        ("_vecs_graph_builder", _vecs_graph),
-        ("_structs_graph_builder", _structs_graph),
+        # ("_vecs_graph_builder", _vecs_graph),
```
acl-cqc, line 167: Hmmm, these are a bit sad :(

acl-cqc (approved):
Looks good, wonder if the commented-out tests can be reinstated for 0.7.1

PR #20, opened 2024-12-16
build: Add a release workflow

johnchildren:
Generated with cargo-dist


johnchildren:
This only works for Rust so far, so will need some tweaking

PR #24, opened 2025-02-25
Experiment/recursive controller

mwpb:
Create an ephemeral controller to run Tierkreis graphs. It can run a modified version of the sample graph and the QSCI time evolution graph on Fugaku.

* Create `examples/numerical-worker` that takes inputs and reads outputs from the filesystem. Create a launcher in `examples/launchers` that launches the numerical-worker.
* Create the `NodeLocation` model for identifying a particular run of a node in a graph.
* Test that the controller works by running the existing sample graph but with match statements removed. The controller does not support match statements as presently implemented because it would have to analyse the values contained therein. Loops are determined to break or continue using a struct instead.
* Create `storage` interface for managing the progress data of the graph. Create draft `executor` interface for running individual nodes.
* The `tierkreis.controller.start:start` function executes a node and then exits. It also takes care of finishing some nodes that don't require computation. The `tierkreis.controller.resume:resume` function recursively analyses the graph trying to find nodes that can be started. At the moment we just immediately execute startable nodes as we find them.
* (Try to get CI to pass.)


python/examples/numerical-worker/__main__.py
```diff
@@ -0,0 +1,98 @@
...
+            a = json.loads(fh.read())
+        with open(node_definition.inputs["b"], "rb") as fh:
+            b = json.loads(fh.read())
+
+        c = a and b
+
+        with open(node_definition.outputs["value"], "w+") as fh:
+            fh.write(json.dumps(c))
+
+    elif node_definition.function_name == "id":
```
johnchildren, line 81: Would it make sense for `id` to be a builtin so you can make use of the linking functionality? Similarly with `copy`

python/tierkreis/controller/executor/shell_executor.py
```diff
@@ -0,0 +1,26 @@
...
+        if not launcher_path.exists():
+            raise TierkreisError(f"Launcher not found: {launcher_name}.")
+
+        if not launcher_path.is_file():
+            raise TierkreisError(f"Expected launcher file. Found: {launcher_path}.")
+
+        with open(self.std_err_path, "a") as fh:
+            subprocess.Popen(
+                [f"{self.launchers_path}/{launcher_name}", node_definition_path],
+                start_new_session=True,
```
johnchildren, line 23: what does `start_new_session` do in this context? I can't quite tell what it is meant to do from the python docs.

python/tierkreis/controller/executor/shell_executor.py
```diff
@@ -0,0 +1,26 @@
...
+    def run(self, launcher_name: str, node_definition_path: Path) -> None:
+        launcher_path = self.launchers_path / launcher_name
+        if not launcher_path.exists():
+            raise TierkreisError(f"Launcher not found: {launcher_name}.")
+
+        if not launcher_path.is_file():
+            raise TierkreisError(f"Expected launcher file. Found: {launcher_path}.")
+
+        with open(self.std_err_path, "a") as fh:
+            subprocess.Popen(
```
johnchildren, line 21: Is there a reason to use `Popen` rather than `run` here? Is it to do with `start_new_session` ?

python/tierkreis/controller/models.py
```diff
@@ -0,0 +1,99 @@
...
+    done_path: Path
+    logs_path: Optional[Path]
+
+
+class NodeType(Enum):
+    NODE = 0
+    LOOP = 1
+    MAP = 2
+
+    def __str__(self) -> str:
```
johnchildren, line 27: Is there a reason to prefer this method over subclassing `StrEnum` ?

python/tierkreis/controller/models.py
```diff
@@ -0,0 +1,99 @@
...
+        return f"{self.node_type}{self.idx}"
+
+    @staticmethod
+    def from_str(frame: str) -> "NodeStep":
+        try:
+            c, idx = frame[0], int(frame[1:])
+            return NodeStep(node_type=NodeType.from_str(c), idx=idx)
+        except (IndexError, ValueError) as exc:
+            logger.exception(exc)
+            raise TierkreisError(f"Invalid NodeStep {frame}")
```
johnchildren, line 63: I might include a `from exc` on this raise, but I'm not certain how much it matters give than you are already logging the exception. I don't know if `logger.exception` will include the stack trace from the original exception and what your intention is there.

python/examples/numerical-worker/__main__.py
johnchildren: I guess there isn't much use in `id` usually

python/tierkreis/pyruntime/python_runtime.py
```diff
@@ -102,6 +102,8 @@ async def run_graph(
         total_nodes = run_g._graph.number_of_nodes()
         runtime_state: dict["_EdgeData", TierkreisValue] = {}
 
+        run_g._graph.add_edge()
```
johnchildren, line 105: Is this intentional?

python/tierkreis/controller/storage/filestorage.py
```diff
@@ -0,0 +1,104 @@
...
+        path = self.workflow_dir / str(node_location) / "outputs" / port_name
+        path.parent.mkdir(parents=True, exist_ok=True)
+        return path
+
+    def _done_path(self, node_location: NodeLocation) -> Path:
+        path = self.workflow_dir / str(node_location) / "_done"
+        path.parent.mkdir(parents=True, exist_ok=True)
+        return path
+
+    def clean_graph_files(self) -> None:
```
johnchildren, line 33: Am I right in thinking this does the "cleanup" by moving it to a random uuid temporary directory? Is there any way to figure out where things were moved after the fact if you have a few directories like this?

python/examples/numerical-worker/__main__.py
mwpb: Yeah I just needed `id` because it is in the sample graph. I don't expect this to get used in real applications.

python/tierkreis/controller/executor/shell_executor.py
mwpb: Not an expert at this so happy to hear other suggestions. (Maybe using `multiprocessing`??) The current approach strings together two man pages. First

https://docs.python.org/3.9/library/subprocess.html#popen-constructor

```
If start_new_session is true the setsid() system call will be made in the child process prior to the execution of the subprocess. (POSIX only)
```

second

https://man7.org/linux/man-pages/man2/setsid.2.html

```
The calling process also becomes the process group leader of a new process group in the session
```

which indicates to me that signals from the controller process will not reach the `subprocess.Popen` process. Some manual testing confirms that when the controller process goes down the `subprocess.Popen` process says up.

python/tierkreis/controller/executor/shell_executor.py
mwpb: Yup, using `run` I couldn't figure out how to get the worker process to stay up when the controller process exits.

python/tierkreis/controller/models.py
mwpb: No! Done.

python/tierkreis/controller/models.py
mwpb: No your suggestion is better! The intention was simply to ease error handling from calling code. Done.

python/tierkreis/pyruntime/python_runtime.py
mwpb: No good catch. I've removed.

python/tierkreis/controller/storage/filestorage.py
mwpb: Hmm... maybe we could change the default checkpoints directory to `Path.home() / ".tierkreis" / "checkpoints"` and use `f"/tmp/tierkreis/archive/{self.workflow_id}/{time_ns()}"` for archived checkpoints. I've pushed some example code. Open to suggestions about other setups also though!

python/tierkreis/controller/models.py
mwpb: Ah now I'm working through the CI I remember... StrEnum was introduced in Python 3.11 and this project still supports Python 3.10. Have reverted for now. In future we may want to consider bumping the Python version.

johnchildren (approved):

PR #25, opened 2025-04-10
Test/new graphdata

mwpb:
Create new GraphData model. Will be a breaking change to tierkreis-chemistry-demo. Not going to the the final internal implementation but after this we should keep breaking changes in the graph builder to a minimum until after the next demo.


python/tierkreis/controller/data/graph.py
```diff
@@ -0,0 +1,80 @@
...
+NodeIndex = int
+OutputLoc = tuple[NodeIndex, PortID]
+ValueRef = tuple[NodeIndex, PortID]
+
+
+@dataclass
+class Func:
+    function_name: str
+    inputs: dict[PortID, ValueRef]
+    type: Literal["function"] = field(default_factory=lambda: "function")
```
johnchildren, line 18: would `field(default="function")` have also worked here instead of the lambda?

python/tierkreis/controller/storage/walk.py
```diff
@@ -92,25 +91,21 @@ def walk_loop(storage: ControllerStorage, node_location: NodeLocation) -> WalkRe
...
-    inputs[Labels.VALUE] = (old_loc, old_port)
-    inputs[Labels.THUNK] = (new_location.append_node(0), Labels.THUNK)
+    inputs = {k: (new_location.append_node(-1), v.name) for k, v in input_paths.items()}
+    inputs[Labels.VALUE] = (new_location, Labels.VALUE)
+    inputs[Labels.THUNK] = (new_location.append_node(-1), Labels.THUNK)
 
     node_run_data = NodeRunData(
         node_location.append_loop(i + 1),
-        FunctionNode(FunctionName("eval")),
+        Eval((0, Labels.THUNK), {}),  # TODO: put inputs in Eval
```
johnchildren, line 108: Is this TODO for a later PR?

python/tierkreis/controller/storage/walk.py
```diff
@@ -43,13 +41,14 @@ def walk_node(storage: ControllerStorage, node_location: NodeLocation) -> WalkRe
 
 
 def walk_eval(storage: ControllerStorage, node_location: NodeLocation) -> WalkResult:
+    logger.debug("walk_eval")
     walk_result = WalkResult([], [])
 
-    message = storage.read_output(node_location.append_node(0), Labels.THUNK)
-    graph = TierkreisGraph.from_proto(Graph.FromString(message))
+    message = storage.read_output(node_location.append_node(-1), Labels.THUNK)
+    graph = GraphData(**json.loads(message))
```
johnchildren, line 48: I'm slightly surprised this works, does it actually work recursively? I wouldn't have expected `nodes` to actually be an instance of `NodeDef` with the default dataclasses library as I didn't think it did validation and construction of child classes.

PR #32, opened 2025-05-02
chore: Add dependabot config

johnchildren:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "uv"
+    directory: "/"
+    insecure-external-code-execution: allow
```
mwpb, line 5: Quick double check: was this to access private registries? After we clean up the Python deps, do we still need this?

mwpb (approved):
One question, not urgent.

.github/dependabot.yml
johnchildren: Fixed!

mwpb: Okay, I'm still not completely clear about what the flag does... but I suppose the best thing is to try with the secure setting and see if it works?

mwpb (approved):

PR #48, opened 2025-05-06
feat: Add worker shims

johnchildren:
I think I've got the shims working in a reasonable way now that passes CI (and the examples all run).

It also means that all workers that use the shims will report their errors automatically.

I think there are a couple of issues with it though:
It assumes json serialisation (I think this could be made configurable)
The shims are part of the main tierkreis dependency (maybe it could be factored out into a tierkreis-worker package?)
"glob" style inputs and outputs rely on special behaviour if there is exactly one input whose type is Iterator or an Iterator output which is slightly magical

Otherwise I think the PR is now ready for review


tierkreis/tierkreis/worker.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+Params = ParamSpec("Params")
+ReturnType = TypeVar("ReturnType", bound=BaseModel)
+
+
+def _load_args(inputs: dict[str, Path]) -> dict:
+    kwargs = {}
+    for arg_name, path in inputs.items():
+        if arg_name not in inputs:
```
mwpb, line 33: Might be reading this wrong... but is this possible?

tierkreis/tierkreis/worker.py
```diff
@@ -0,0 +1,161 @@
...
+            func_name = name if name is not None else func.__name__
+
+            def wrapper(node_definition: WorkerCallArgs):
+                annotations = func.__annotations__
+                annotations.pop("return")
+                # If there is only one argument and its value is iterator,
+                # we need to load multiple files to the iterator.
+                if (
+                    len(annotations) == 1
+                    and get_origin(next(iter(annotations.values()))) is Iterator
```
mwpb, line 113: Yes indeed this is getting a little complicated! Not so much a suggestion for this PR but something to think about for the future: can we restrict use of globs to the builtins worker (where they could just be implemented by manual file management) and encourage users to use higher level functions like fold, filter etc...

tierkreis/tierkreis/worker.py
```diff
@@ -0,0 +1,161 @@
...
+                        input_iterator_name, node_definition.inputs
+                    )
+                    results = func(iterator)
+                else:
+                    kwargs = _load_args(node_definition.inputs)
+                    results = func(**kwargs)
+
+                if isinstance(results, Iterator):
+                    _save_results_iterator(node_definition.output_dir, results)
+                else:
```
mwpb, line 126: Maybe one way around the JSON assumption is to deal with non-JSONable types here. I.e. if we see a `bytes` annotation just return the raw bytes.

mwpb (approved):
Thanks John, looks good for now (and I suppose we want to iterate a bit before end-users get their hands on it?).

* Is the intention that non-JSON inputs should just be dealt with manually as we did previously?
* In future I'm definitely keen to try to avoid end-users having to deal with the glob patterns. Might need a bit of design work first.

johnchildren:
> Thanks John, looks good for now (and I suppose we want to iterate a bit before end-users get their hands on it

I think so, it's still quite rough and I think there are probably a few low hanging improvements. For now I'm mostly keen to just get it working minimally as having the error handling by default is quite useful for authoring and debugging workflows.

>     * Is the intention that non-JSON inputs should just be dealt with manually as we did previously?

I have a couple of alternative ideas here, let me know if you have a preference:

* We could make the user provide a serialiser/deserialiser when instantiating `Worker` like:
  ```python
  worker = Worker("my-worker", serialiser=json.dumps, deserialiser=msgpack.loads)
  ```
  Which could also be override on a per function level
* We add more special types that allow a lower level control, like
  ```python
    @worker.function(output_files={"out"})
    def my_special_function(*, a: File, out: File):
        pass
  ```
  Where the "out" file is opened in write mode.

>     * In future I'm definitely keen to try to avoid end-users having to deal with the glob patterns. Might need a bit of design work first.

Yeah, it's worth thinking about. I do actually kind of like the iterator-y access mode but it's a bit too complicated for regular users perhaps.

tierkreis/tierkreis/worker.py
johnchildren: Ah yes, sorry I didn't read this before commenting but that would make sense here I think.

mwpb:
All three seem reasonable, although for the second we might have to abstract the storage interface a bit.

PR #49, opened 2025-05-06
Feat/error reporting

philipp-seitz:
*Summary:* This adds minimal capabilities to report errors in function workers.

*Details*
- Adds a new state in the frontend: `Error` which is currently not propagated up through nodes
- Changes the contract for workers: The `WorkerCallArgs` now include an `error_path` argument
- Workers now are responsible for creating the file in `error_path`
- Executors now will direct `stderr` to an `errors` file which is read by the frontend for detailed information

*Potential Issues*
- If a worker fails due to an exception e.g. oom, this is not handled yet


mwpb (commented):
Thanks Philipp! This looks good in general. In particular it successfully brings the error all the way from the worker to the visualiser. My main suggestion would be to include the errored nodes in the `WalkResult`. If we do this then we would be able to check for them on the next tick of the controller and print out the appropriate error. In the first instance we should probably stop the controller at that point also.

mwpb (approved):
Thanks for the changes. This looks good to me. (Have tested locally and read the code.)

python/tierkreis/controller/__init__.py
```diff
@@ -41,6 +41,9 @@ def resume_graph(
 ) -> None:
     for i in range(n_iterations):
         walk_results = walk_node(storage, Loc())
+        if walk_results.errored != []:
+            print("Graph finished with errors.")
```
johnchildren, line 45: I wonder if it's worth using a logger here and logging at `error` level so it looks a bit more serious than a print. Very minor though and could be a follow up PR.

johnchildren: Ah I guess you are already doing that when you read the error?

johnchildren (approved):
I'm not completely convinced about having a separate file for error logs versus regular logs, but we can discuss / revisit if needed. Otherwise changes look good. I had a small comment about use of print versus logging but it's very minor and can be ignored

python/tierkreis/controller/__init__.py
mwpb: Yes I agree this should be improved. The logs configuration is a pre-existing issue though so I've made a separate Jira ticket for it.

PR #60, opened 2025-05-08
Feature/error-visualization

philipp-seitz:
In the visualizer, errors in nodes are know checked for in children.

On error exit, the controller will now write a file with all the node locations containing an error to the error_logs of the toplevel node.


python/tests/errors/test_error.py
```diff
@@ -26,6 +29,14 @@ def wont_fail() -> GraphData:
     return graph
 
 
+def fail_in_evla() -> GraphData:
```
johnchildren, line 32: typo?

Issue #64, opened 2025-05-08
Viewing a node in the visualiser that hasn't started yet results in a 500 error


johnchildren:
![Image](https://github.com/user-attachments/assets/4a7af794-861c-4b78-9023-c9b29c7e0f7d)

Clicking one of these white nodes results in a 500 internal error, it should probably be at least a 404 or maybe a page with the node definition alone.

See the below trace for the exact exception:

```
Traceback (most recent call last):
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/uvicorn/protocols/http/httptools_impl.py", line 409, in run_asgi
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/fastapi/applications.py", line 1054, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/applications.py", line 112, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 187, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/middleware/exceptions.py", line 62, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/routing.py", line 714, in __call__
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/routing.py", line 734, in app
  File "/home/john/Dojo/tierkreis/.devenv/state/venv/lib/python3.12/site-packages/starlette/routing.py", line 288, in handle
[...]

PR #70, opened 2025-05-09
Fix/404nodes

philipp-seitz:
fixes: #64

I've created error handling in case a node has no associated data.

Screenshot 2025-05-09 at 10 12 20

### DIFF
--- a/tierkreis_visualization/tierkreis_visualization/routers/workflows.py
+++ b/tierkreis_visualization/tierkreis_visualization/routers/workflows.py
@@ -7,7 +7,7 @@ from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
 from starlette.responses import JSONResponse, PlainTextResponse
 from tierkreis.controller.data.location import Loc, WorkerCallArgs
-from watchfiles import awatch
+from watchfiles import awatch, Change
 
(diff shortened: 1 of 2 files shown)


tierkreis_visualization/tierkreis_visualization/routers/workflows.py
```diff
@@ -88,6 +97,11 @@ async def node_stream(workflow_id: UUID, node_location: Loc):
     metadata_path = (
         CONFIG.tierkreis_path / str(workflow_id) / str(node_location) / "_metadata"
     )
+    if not metadata_path.exists():
+        async for _changes in awatch(metadata_path.parent):
+            if _changes[0] == Change.added:
+                break
```
philipp-seitz, line 103: I'm not sure if this is working as intended, is there a good way to test this @johnchildren

johnchildren (approved):

tierkreis_visualization/tierkreis_visualization/routers/workflows.py
johnchildren: I'm not really sure, I don't know what conditions cause the metadata file to not be written yet.

Issue #58, opened 2025-05-08
Add a way to run a graph in an "easy" way.


johnchildren:
Currently there is a bit of ceremony and configuration required to run a workflow, it would be nice if there was a way to run a workflow with minimal setup for first time users. I could imagine there being something like:

```
outputs = run_easy_workflow(my_graph(), {"input_foo": 3})
```

Which manages the setup for the executor and the storage with sensible defaults and automatically restarts the graph if checkpoints already exist.

Issue #72, opened 2025-05-09
Make it easy to restart a workflow by default


johnchildren:
When running a workflow that has started previously with `run_graph` you will get an error due to files already existing. It would be useful if there was an entrypoint that just tried to resume the graph. This was achieved in the chemistry demo by passing an explicit flag to restart the workflow, but it would be preferable if this just happened automatically in most cases.

```python
    if not restart_only:
        storage.clean_graph_files()
        inputs = {"molecular_inputs": json.dumps(EXAMPLE_INPUT).encode()}
        run_graph(storage, executor, g, inputs, polling_interval_seconds=1)
    else:
        resume_graph(storage, executor, polling_interval_seconds=1)
```

```
❯ uv run examples/symbolic_circuits/symbolic_circuits.py
Starting workflow at location: /home/john/.tierkreis/checkpoints/00000000-0000-0000-0000-000000000000/logs
Traceback (most recent call last):
  File "/home/john/Dojo/tierkreis/examples/symbolic_circuits/symbolic_circuits.py", line 116, in <module>
    main()
  File "/home/john/Dojo/tierkreis/examples/symbolic_circuits/symbolic_circuits.py", line 99, in main
    run_graph(
  File "/home/john/.cache/uv/environments-v2/symbolic-circuits-7d33ca96fd86a47c/lib/python3.12/site-packages/tierkreis/controller/__init__.py", line 35, in run_graph
    start(storage, executor, node_run_data)
  File "/home/john/.cache/uv/environments-v2/symbolic-circuits-7d33ca96fd86a47c/lib/python3.12/site-packages/tierkreis/controller/start.py", line 108, in start
    pipe_inputs_to_output_location(storage, node_location.N(-1), ins)
  File "/home/john/.cache/uv/environments-v2/symbolic-circuits-7d33ca96fd86a47c/lib/python3.12/site-packages/tierkreis/controller/start.py", line 153, in pipe_inputs_to_output_location
    storage.link_outputs(output_loc, new_port, old_loc, old_port)
  File "/home/john/.cache/uv/environments-v2/symbolic-circuits-7d33ca96fd86a47c/lib/python3.12/site-packages/tierkreis/controller/storage/filestorage.py", line 135, in link_outputs
    os.link(self._output_path(old_location, old_port), new_dir)
[...]

Issue #59, opened 2025-05-08
Add a way to run the controller from the CLI


johnchildren:
Similar to https://github.com/CQCL/tierkreis/issues/58

It would be nice if there was a way to run the controller as a CLI utility with file arguments, this way you could do something like:

```
tkr graph_module:my_graph --input-foo foo.json --input-bar bar.json --output-baz baz.json
```

Where the command line arguments specify local files that are either copied or linked into/out of the tierkreis store.

PR #75, opened 2025-05-09
Feat: cli and graph restarting

philipp-seitz:
I've accumulated a couple of QoL improvements in this:
Closes: #58 #59 #72

In summary:
- added a cli, although we still need a convenient way to provide a graph to it
- added a `run_workflow` wrapper
- added two flags to the storage controller to prevent the issue of already existing graphs


tierkreis/tierkreis/controller/__init__.py
```diff
@@ -21,6 +21,7 @@ def run_graph(
     graph_inputs: dict[str, bytes],
     n_iterations: int = 10000,
     polling_interval_seconds: float = 0.01,
+    **_kwargs,
```
mwpb, line 24: Would have a preference for keeping going with named arguments here.

mwpb (commented):
Hi, thanks Philipp, some code comments follow. My initial reaction is that this might be a nice way to start simple workflows locally. However if it is intended as the main way to run workflows we might have to think a little harder about how it interacts with the executors etc..

tierkreis/tierkreis/cli/run_workflow.py
```diff
@@ -0,0 +1,56 @@
...
+        graph,
+        inputs,
+        **{
+            k: v
+            for k, v in kwargs.items()
+            if k in ["n_iterations", "polling_interval_seconds"]
+        },
+    )
+    if print_output:
+        output = json.loads(storage.read_output(Loc(), Labels.VALUE))
```
mwpb, line 55: Just noticed this. I reckon that `print_output` should be an Optional PortID instead of a bool.

mwpb: Or programmatically get the outputs.

tierkreis/tierkreis/cli/run_workflow.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    if run_id is None:
+        workflow_id = UUID()
+    else:
+        workflow_id = UUID(int=run_id)
+    storage = ControllerFileStorage(workflow_id, name=name)
+    if registry_path is None:
+        registry_path = Path(__file__).parent
+    executor = UvExecutor(registry_path=registry_path, logs_path=storage.logs_path)
```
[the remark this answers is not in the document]
mwpb: For the record, run workflow is intended to be a 'quick' way to run simple workflows so we will use ShellExecutor. For configuration of executors we will drop down to `run_graph`.

mwpb (approved):

README.md
```diff
@@ -3,6 +3,23 @@
 Quantum-classical hybrid workflow orchestration tool.
 This is the top level repository containing the python packages.
 
+## CLI
+
+Tierkreis comes with a command line interface for running workflows.
+To see all available options use `tkr --help`.
+To run the hello world example from the cli
+```
+uv run tkr -g examples/hello_world/hello_world_graph.py:hello_graph --i data.json --uv --registry-path examples/hello_world/ -o.
```
johnchildren, line 12: Is the dot at the end of this line intentional?

johnchildren (approved):

johnchildren (approved):

PR #86, opened 2025-05-13
test: cli

philipp-seitz:


tierkreis/tests/cli/test_run_workflow.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.fixture()
+def graph() -> GraphData:
+    return simple_eval()
+
+
+def test_run_workflow(graph: GraphData) -> None:
+    inputs = {}
+    run_workflow(inputs=inputs, graph=graph, run_id=31415)
+    with open(
+        Path.home()
```
johnchildren, line 25: I guess our other tests are also using the default directory, but it would be nice if we were to use a temporary directory. Previously I have used tempfile for this kind of thing: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

johnchildren (approved):

tierkreis/tests/cli/test_run_workflow.py
philipp-seitz: I was thinking about this, the problem is that the storage is inaccessible through `run_workflow` and we can not overwrite it

PR #87, opened 2025-05-13
chore: Move common workers into a package

johnchildren:
Starts adding a new package for common workers and refactors the examples folder a little to start using multiple executors.


mwpb (approved):

tierkreis_workers/tierkreis_workers/__init__.py
```diff
@@ -0,0 +1,4 @@
+from pathlib import Path
```
johnchildren, line 1: Actually this isn't very useful as it introduces a dependency perhaps?

PR #88, opened 2025-05-13
chore: Further refacotring of workers 

johnchildren:
Move in some of the chemistry demo workers and start further refactoring and packaging. Also reduces the amount of tupling in the hamiltonian example

### DIFF
--- a/tierkreis/tierkreis/worker.py
+++ b/tierkreis/tierkreis/worker.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from collections.abc import (
     Iterator,
 )  # Needs to be imported from here not typing to satisfy get_origin
+import sys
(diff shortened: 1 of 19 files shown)


tierkreis/tierkreis/worker.py
```diff
@@ -78,6 +79,13 @@ def __init__(self, name: str) -> None:
         self.name = name
         self.functions = {}
 
+        def handle_unhandled_exception(exc_type, exc_value, exc_traceback):
+            logger.critical(
+                "Unhandled exception", exc_info=(exc_type, exc_value, exc_traceback)
+            )
+
+        sys.excepthook = handle_unhandled_exception
```
mwpb, line 87: Certainly glad we don't have to write this every time... :).

mwpb (approved):

PR #90, opened 2025-05-14
Feat: eagerifelse

philipp-seitz:
This adds an `eifelse` statement that evaluates eagerly


tierkreis/tests/controller/test_eagerifelse.py
```diff
@@ -0,0 +1,49 @@
...
+    out = g.add(EagerIfElse(pred_long, one_long, two))(Labels.VALUE)
+    g.add(Output({"simple_eagerifelse_output": out}))
+    return g
+
+
+params = [({"pred": b"true"}, 1), ({"pred": b"false"}, 2)]
+
+
+@pytest.mark.parametrize("input, output", params)
+def test_eagerifelse_long_running(input: dict[str, bytes], output: int) -> None:
```
mwpb, line 39: This looks like it runs by default. If that is the case then it would be better to [skip long running tests](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option) by default.

tierkreis/tests/controller/test_resume.py
```diff
@@ -45,6 +47,7 @@
     "factorial_4",
     "factorial_5",
     "loop_multiple_acc",
+    "simple_eagerifelse",
```
mwpb, line 50: Hmm... now I think about it, it might be nice to have a separate test just for both IfElse implementations. For the eager version we can look into the file system and see that both branches have produced outputs. For the lazy one check that only one branch has produced outputs. Probably I should have done this in the PR that introduced the lazy IfElse, as that one has the unusual behaviour... Not strictly speaking required for this PR.

tierkreis/tierkreis/controller/storage/walk.py
```diff
@@ -163,3 +167,22 @@ def walk_map(
...
+
+def walk_eifelse(
+    storage: ControllerStorage,
+    parent: Loc,
+    idx: NodeIndex,
+    node: EagerIfElse,
+) -> WalkResult:
+    loc = parent.N(idx)
+    children = [node.if_true, node.if_false]
+    if all(storage.is_node_finished(parent.N(child[0])) for child in children):
```
mwpb, line 180: Is this still required?

mwpb (commented):
Thanks Philipp, in general looks good. I think you can simplify your `walk_ifelse` though because the condition is already checked in the `in_edges`?

philipp-seitz:
@mwpb I addressed your comments, what do you think about the `main.py` in the controller? I used this out of laziness

mwpb (approved):
Thanks those changes look good. The `tests/controller/main.py` is a bit odd but not a huge problem. We can separate out a proper worker for tests later if we want.

PR #93, opened 2025-05-15
feat: implement Partial

mwpb:
Simple implementation of partial graphs by inserting `Loc` data into the `GraphData`. In future, if we need more performance we can avoid the repeated deserialisation of the GraphData by passing the fixed inputs through in the NodeRunData at the cost of higher code complexity. The diff looks worse than it is because I had to extract out some variables to avoid Python circular imports.


tierkreis/tierkreis/controller/data/graph.py
```diff
@@ -24,6 +24,13 @@ class Eval:
     type: Literal["eval"] = field(default="eval")
 
 
+@dataclass
+class Partial:
```
johnchildren, line 28: This difference between Partial and Eval is just that Partial _can_ return a graph if not all of the input ports are satisfied right? Would it not be easier to just have a single operation that does both as they both just depend on the graph and how many inputs are provided?

mwpb: Sure, this is possible. I'm a little worried that would cause confusion when someone missed an argument. Though I suppose it doesn't seem to be confusing in ML style with something like `let add1 = add 1`. I'll have a think about implementation details; we'd need to look up all the input nodes in a graph at least.

mwpb: Okay have swapped over to using `Eval` for both now.

tierkreis/tierkreis/controller/data/graph.py
```diff
@@ -86,6 +86,8 @@ class Output:
 
 class GraphData(BaseModel):
     nodes: list[NodeDef] = []
+    fixed_inputs: dict[PortID, OutputLoc] = {}
+    inputs: set[PortID] = set()
```
johnchildren, line 90: I am a bit confused about why `inputs` and `outputs` have different types, is it because `outputs` is all the outputs of nodes in the graph and `inputs` is the inputs to the graph?

I wonder also if you could combine both of these types to just be:

`dict[PortID, OutputLoc | None]`

To avoid needing to do the set subtraction

```python
def remaining_inputs(self, provided_inputs: set[PortID]) -> set[PortID]:
    # check for inputs that are already set
    # ...


    return {input for input, maybe_bound in self.inputs
                    if maybe_bound is None and input not in provided_inputs}
```

tierkreis/tierkreis/controller/start.py
```diff
@@ -104,7 +113,17 @@ def start(
         storage.mark_node_finished(node_location)
 
     elif node.type == "eval":
+        message = storage.read_output(parent.N(node.graph[0]), node.graph[1])
+        g = GraphData(**json.loads(message))
```
johnchildren, line 117: I wondered if there was a "better" way to do this to avoid reading out and modifying the graph but I doubt that there is an easy alternative.

The only other thing I would think of is perhaps having an auxiliary file that contains the "fixed_inputs" values that can be copied alongside the graph file. That way you could potentially keep the graph symlinked and just modify the fixed inputs file if more are added. But this does just add complexity elsewhere.

johnchildren (approved):
Looks good overall I think, but I just had a couple of comments that you can ignore but could be worth thinking about.

tierkreis/tierkreis/controller/start.py
johnchildren: I guess this notion is very similar to "context" in interpreters

tierkreis/tierkreis/controller/data/graph.py
mwpb: This is essentially correct. I have renamed the variables `inputs` -> `graph_inputs` and `outputs` -> `node_outputs` to make things easier to read. Disregarding performance considerations, I find it a bit more readable to use `set` rather than add `None` in the codomain. Am happy to change over if you'd like though.

tierkreis/tierkreis/controller/start.py
mwpb: Yes that makes sense, I wanted to avoid complexity in that way at the beginning. Especially as graphs are meant to be small enough to load into memory. From the perspective of object algebras, I think this corresponds to extra attributes of the class implementing Eval that stores these inputs. E.g. an attribute of `AddExprEval` [here](http://ponies.io/posts/2015-07-15-solving-the-expression-problem-in-python-object-algebras-and-mypy-static-types.html). However as you say we are going to have to store them somewhere!

The natural way to avoid reading the `GraphData` in start would be to only read it in `walk` (where we have to anyway) and then pass the fixed inputs as a new attribute of `NodeRunData`.

PR #98, opened 2025-05-15
Add hpc executor draft

### LINKED ISSUE #83: Add an example or re-usable hpc executor for use with pjsub
For the chemistry project we developed an internal executor that supported the use of the fujitsu queuing system by invoking pjsub. As we are targeting HPC systems with this project we should provide either an example of how to do it or an executor implementation that can be used by regular users.

philipp-seitz:
closes #83


philipp-seitz:
closed in favor of other proposal

PR #99, opened 2025-05-16
Docs/readme

philipp-seitz:
closes: #69 
Update of the tierkreis package readme


johnchildren:
This looks good, thank you. I may even use your examples for the graph and setting up the runtime in my presentation. I think it's a bit more straightforward than the hello world graph as that requires setting up the uv executor to run the custom worker.

johnchildren (approved):

PR #100, opened 2025-05-19
feat: Add tierkreis TUI

johnchildren:
This is essentially just a log viewer for the checkpoints folder that lets you find the logs file a bit easier and will also render an error message if any in a red box.


johnchildren:
image
image
image
example screenshots

johnchildren:
image
image

johnchildren:
> just a heads up I made a slight change to the workers which fits what I did in the visualizer:
> - the `_error` file should work like `_done`, i.e. it's only a flag
> - error logs should go to `errors` in the node directory
> We made this choice since we also pass `errors` as a handle to the subprocess as stderr which means it will always be created compared to `_error`

Fair enough, I think the worker shims are probably writing to the wrong location in that case

PR #103, opened 2025-05-20
fix: use NodeDef for is_node_started

mwpb:
Then we can avoid writing the worker call args for every node.


mwpb:
Code is too stale; I'll start a new branch.

PR #108, opened 2025-05-21
fix: vis for ifelse

mwpb:
Use dashed line for conditional edges that haven't evaluated. (Currently only in lazy `IfElse`.)

* (Fix pyright errors.)
* Add the conditional edges for IfElse in the Eval visualisation. (Make some variable names more consistent.)
* Use the jinja built-in `tojson` instead of safe when we want to put JSON in the HTML.


mwpb:
Screenshot 2025-05-22 at 09 45 43
Screenshot 2025-05-22 at 09 45 53

aidanCQ (approved):

PR #113, opened 2025-05-22
Draft simpler builder interface

philipp-seitz:
I've created a draft for a simplified builder interface that lets you chain nodes.
I made a minimal working example, the `Graphdata` is not completely done.
I provided some defaults, e.g. `output()` will consume the port `Labels.VALUE` from the previously added node.
For multiple inputs, you can provide a `dict[str, int]` where the values are the index of the node in the chain to add.
I'm not super happy with this solution so open to suggestions.
As soon as you have multiple outputs it can get quite confusing, as the defaults no longer make sense.

An alternative approach would be similar to `networkx` where you provide all the nodes and edges upfront.
For this approach, I couldn't find a good way to provide the port information.
you could have something like this though:

```python
nodes =  [Const(1),Const(2),Func("builtins.iadd"),Output()]
edges = [(0,2,"a"),(1,2,"b"),(2,3)] # without label use "value"
```

### DIFF
--- a/examples/builder.py
+++ b/examples/builder.py
@@ -1,0 +1,30 @@
+from pathlib import Path
+from tierkreis.cli.run_workflow import run_workflow
+
+from tierkreis.controller.data.graph import GraphData
+from tierkreis import Labels
+
+
+def loop_body() -> GraphData:
+    g = GraphData()
+    g.input().const(1).const(10).func("builtins.iadd", {"a": 0, "b": 1}).func(
+        "builtins.igt", {"a": 2, "b": 3}
+    ).output({Labels.VALUE: -2, "should_continue": -1})
+    return g
+
+
+f = GraphData()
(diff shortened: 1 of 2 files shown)


examples/builder.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+from tierkreis.cli.run_workflow import run_workflow
+
+from tierkreis.controller.data.graph import GraphData
+from tierkreis.controller.data.graph_builder import AutoGraphBuilder, GraphBuilder
+from tierkreis import Labels
+
+
+def loop_body() -> GraphData:
```
johnchildren, line 9: I'm pretty unconvinced by this approach, while it's less verbose it seems like it would be very awkward to make changes as adding new nodes in between existing nodes will mean updating every index. It's also very error prone to work this way as you are very vulnerable to off by one errors.

I guess if you have some degree of validation you might be able to avoid some issues, but I'm not sure this is very workable as an approach.

PR #116, opened 2025-05-27
Experimental: new visualizer

philipp-seitz:
Draft for a new node visulaizer based on `bun` and `reactflow`


tierkreis_visualization/frontend/src/components/ui/card.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from "react"
+
+import { cn } from "@/lib/utils"
```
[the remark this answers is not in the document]
philipp-seitz: yes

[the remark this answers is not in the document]
mwpb: Good, in this case then let's just add an override in a local .gitignore so that we version control this one `lib` folder.

tierkreis_visualization/tierkreis_visualization/data/loop.py
```diff
@@ -20,19 +21,44 @@ def get_loop_node(
...
-
+    nodes.append(
+        PyNode(
+            id=i, status=last_status, function_name=f"L{i}", node_location=new_location
+        )
+    )
+    edges = []
+    for port_name in storage.read_worker_call_args(node_location.L(0)).inputs:
+        if port_name not in storage.read_output_ports(new_location):
+            print(port_name, node_location)
```
johnchildren, line 48: Do you still need this print?

tierkreis_visualization/tierkreis_visualization/routers/workflows.py
```diff
@@ -152,3 +200,15 @@ def get_logs(workflow_id: UUID, node_location_str: str):
...
+
+@router.get("/{workflow_id}/nodes/{node_location_str}/errors")
+def get_errors(workflow_id: UUID, node_location_str: str):
+    node_location = parse_node_location(node_location_str)
+    storage = get_storage(workflow_id)
+    if not storage.node_has_error(node_location):
+        return PlainTextResponse("Node has no errors.", status_code=404)
+
+    messages = storage.read_errors(node_location)
+    print(messages)
```
johnchildren, line 213: Do you still need this?

tierkreis_visualization/tierkreis_visualization/main.py
```diff
@@ -10,6 +11,15 @@
 
 app = FastAPI()
 
+app.add_middleware(
```
johnchildren, line 14: I guess this could be necessary now two web servers are involved, but it would be nice to avoid.

philipp-seitz: this can be avoided in a production setting were we server the js as static pages. For development it's convenient so hot reloading works properly without bundling.

PR #117, opened 2025-06-03
feat: generate type safe graphbuilder from worker

mwpb:
Possible way to type check graph creation.

* Workers can generate stubs for the graph builder to use. It cannot handle complex generics or how we are using Iterators yet.
* New graph builder methods like const, eval, etc.. which are statically typed check by pyright. These are not completely watertight (the graph builder could fake some of the generated stubs and go wrong for instance...) but I think should eliminate the most common kinds of errors.
* Problem 1: significant additional complexity in the Tierkreis library code. Only just scratching the surface with the current PR. E.g. we probably should update the `Worker.run` to align more closely.
* Problem 2: where do we draw the line with nested types? I'm not keen to load up arbitrary Python code into the graph builder. One solution would be only to allow things that serialise to a JSON schema or similar... it would add yet another layer of conversions though. Currently this PR just treats types as opaque below a certain level.


tierkreis/tierkreis/namespace.py
```diff
@@ -0,0 +1,64 @@
...
+@dataclass
+class Namespace:
+    name: str
+    functions: list[FunctionSpec]
+
+    def add_from_annotations(self, name: str, annotations: dict[str, Any]) -> None:
+        try:
+            fn = FunctionSpec(name=name, namespace=self.name, ins={}, outs={})
+            fn.add_inputs({k: v for k, v in annotations.items() if k != "result"})
+            fn.add_outputs(annotations["return"])
```
johnchildren, line 59: is annotations here the python __annotations__ or a custom dictionary? I'm slightly unclear. I would have expected "return" to be a 'type' here? Is that correct?

johnchildren (commented):
Sorry for the delay in reviewing this, I have a few comments but I think this is probably a step in the right direction.

I think we should be a bit wary of the TKType type now as we are starting to encode which kinds of values are permissible as TKTypes slightly implicitly, we probably ought to have a spec or a protocol to define this more explicitly in my opinion.

tierkreis/tests/tkr_builtins.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Callable, Literal, NamedTuple
+from tierkreis.controller.data.core import TKRRef, Function, NodeIndex
+
+
+class iadd(Function[TKRRef[int]]):
+    namespace: str = "builtins"
+    a: TKRRef[int]
+    b: TKRRef[int]
+
+    out: Callable[[NodeIndex], TKRRef[int]] = TKRRef[int].from_nodeindex
```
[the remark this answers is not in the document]
mwpb: That would be a good improvement! (Was just moving quite fast at the prototyping stage and didn't spot that possibility.) Will change the codegen...

tierkreis/tierkreis/controller/data/core.py
```diff
@@ -1,7 +1,86 @@
...
+
+class TKRRef[T: TKType](NamedTuple):
+    node_index: NodeIndex
+    port: PortID
+
+    @staticmethod
+    def from_nodeindex(idx: NodeIndex, port: PortID = "value") -> "TKRRef[T]":
+        return TKRRef[T](idx, port)
+
+    def _todict(self) -> dict[str, Any]:
```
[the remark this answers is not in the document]
mwpb: Sure have changes. (Was just imitating `_asdict` from `NamedTuple`.)

tierkreis/tierkreis/namespace.py
mwpb: It is the Python annotations, which pyright says is `dict[str, Any]`. The `annotations["return"]` is indeed intended to be the return type. Now I think more about it, there is the question of a possible `None` return... I'll try and tighten things up in the `FunctionSpec` methods.

mwpb: Okay have tightened things up a little bit and handled `None`.

mwpb:
Thanks John, I agree that we need to think through `TKRType` (renamed for consistency). At least we should figure out:-
* How do we get graph builder stubs for workers that do not use the provided Python `Worker` class decorators?
* How to handle pure `bytes` arguments? Maybe just load the raw bytes from the file?

johnchildren (approved):

PR #125, opened 2025-07-03
Experiment/unwrapped types in graph builder

mwpb:


mwpb:
Closing as there are too many edge cases to keep track of. E.g. what to do if the graph author passes in x+1, where x is a reference? The fundamental difficulty is that the Python interpreter goes first before we get our hands on the values.

PR #135, opened 2025-07-18
fix: map visualization

philipp-seitz:
This fixes the visualization of map nodes by writing an eval node definition for each map element.

### DIFF
--- a/tierkreis/tierkreis/controller/start.py
+++ b/tierkreis/tierkreis/controller/start.py
@@ -150,6 +150,8 @@ def start(
             pipe_inputs_to_output_location(
                 storage, node_location.M(p).N(-1), eval_inputs
             )
+            # Necessary in the node visualization
+            storage.write_node_def(node_location.M(p), Eval((-1, "body"), {}))
 
     elif node.type == "ifelse":
         pass
--- a/tierkreis_visualization/tierkreis_visualization/routers/workflows.py
+++ b/tierkreis_visualization/tierkreis_visualization/routers/workflows.py
@@ -1,4 +1,5 @@
 import json
+import logging
 from typing import Any, assert_never
 from uuid import UUID
 
@@ -22,6 +23,7 @@ from tierkreis_visualization.routers.models import PyGraph
 from tierkreis_visualization.routers.navigation import breadcrumbs
 
 router = APIRouter(prefix="/workflows")
+logger = logging.getLogger(__name__)
 
 
 # Update the routes section with this new route:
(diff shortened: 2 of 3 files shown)


tierkreis/tierkreis/controller/start.py
```diff
@@ -150,6 +150,8 @@ def start(
             pipe_inputs_to_output_location(
                 storage, node_location.M(p).N(-1), eval_inputs
             )
+            # Necessary in the node visualization
+            storage.write_node_def(node_location.M(p), Eval((-1, "body"), {}))
```
mwpb, line 154: This makes good sense. At the moment, we are essentially using the existence of `node_def` as a proxy for 'node has started', with its contents providing useful debug information. I might put the node.inputs in instead of `{}`, even though they are referring to nodes two levels above rather than just one.

tierkreis_visualization/tierkreis_visualization/routers/workflows.py
```diff
@@ -205,16 +207,24 @@ def get_logs(workflow_id: UUID, node_location_str: str):
     storage = get_storage(workflow_id)
     if not storage.is_node_started(node_location):
         return PlainTextResponse("Node not started yet.")
-    definition = storage.read_worker_call_args(node_location)
-
-    if definition.logs_path is None:
+    try:
+        definition = storage.read_node_def(node_location)
```
mwpb, line 211: My preference would be to keep things simple here and insist that only function nodes and the whole workflow can have log files. (We might like to create another endpoint `/{workflow_id}/logs` if the frontend needs to get logs for the whole workflow.) Any other attempt should just get an error. So I think we only really need one call to the storage with `storage.read_worker_call_args`. (Incidentally the existing `storage.is_node_started` looks redundant and can be removed so long as we catch errors from `storage.read_worker_call_args`.)

mwpb (approved):
Thanks Phillip. Looks good in general and should be okay to merge. I would prefer to keep things a bit less forgiving in the logs endpoint so that a caller simply gets an error if they ask for the wrong thing. Not vital though.

tierkreis_visualization/tierkreis_visualization/routers/workflows.py
mwpb: Yes I'd go with the simple version for now. It should be pretty easy to adapt to some future change in requirements, and that change may not ever come anyway.

tierkreis_visualization/frontend/src/App.tsx
```diff
@@ -82,6 +83,48 @@ const Main = (props: {
...
+      if (newEdges.length > 0) {
+        setEdges([...edges, ...newEdges]);
+      }
+    };
+    return () => {
+      if (ws.readyState == WebSocket.OPEN) {
+        ws.close();
+      }
+    };
+  }, [props, reactFlowInstance, setNodes, setEdges]);
```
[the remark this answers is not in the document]
philipp-seitz: ah yes this is not map specific, I just realized when fixing earlier that this was not working properly

tierkreis_visualization/frontend/src/App.tsx
```diff
@@ -141,6 +183,27 @@ export default function App() {
     staleTime: 1000 * 60 * 5, // Cache for 5 minutes
   });
   const workflow_id = workflow_id_url || workflowsQuery.data[0].id;
+
+  const logs = useSuspenseQuery({
+    queryKey: ["workflowLogs", workflow_id],
```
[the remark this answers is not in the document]
mwpb: That's okay for now but we might like to revisit if this gets awkward for the end-user. I'm assuming that the user can just switch to another workflow and back again to see the updated logs? (Watching workflows as they run is definitely in scope!)

PR #142, opened 2025-07-23
chore: Remove type ignore in example

johnchildren:
Removes a type ignore in the symbolic circuits example


mwpb:
This looks good, but unfortunately ruff is annoyed. I'm tempted to ignore ruff F821 (undefined-name) in situations like this; if pyright is happy about undefined names then I probably am also.

mwpb:
Honestly I'd consider an undefined-name error the domain of pyright not ruff so we might even consider disabling undefined-name globally. Indeed in this case pyright realises that the code is correct.

PR #143, opened 2025-07-23
feat: Use PTypes in the inputs to run_graph

johnchildren:
Simplifies the run_graph function a little by converting all ptypes to bytes instead.

### DIFF
--- a/tierkreis/tests/cli/test_tkr.py
+++ b/tierkreis/tests/cli/test_tkr.py
@@ -7,6 +7,7 @@ from uuid import UUID
 
 from tierkreis.cli.tkr import load_graph, _load_inputs, main
 from tierkreis.controller.data.graph import GraphData
+from tierkreis.controller.data.types import PType
 from tierkreis.exceptions import TierkreisError
 
 from tests.controller.sample_graphdata import simple_eval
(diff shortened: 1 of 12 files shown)


tierkreis/tests/cli/test_tkr.py
```diff
@@ -48,15 +49,15 @@ def test_load_graph_invalid() -> None:
             "input1:tierkreis/tests/cli/data/input1",
             "input2:tierkreis/tests/cli/data/input2",
         ],
-        {"input1": b"test", "input2": b'{"a": 5, "b": "string"}'},
+        {"input1": "test", "input2": {"a": 5, "b": "string"}},
```
johnchildren, line 52: Slightly unsure about this change for input1, previously it was raw bytes but I'm not certain how to replicate that behavior without a type annotation

johnchildren: Ah I guess the simplest thing to do for now might be to catch the exception and just ignore it while treating it as bytes. It would be nice if we could read the annotations from the graph itself though, but this might require saving them as part of the graph as a naive approach.

johnchildren: I've updated the PR with your suggestion https://github.com/CQCL/tierkreis/pull/143/commits/ef8d09a9801fccb8a479dfe457c6b8d3ee81428d

mwpb (approved):
Thanks!

Issue #114, opened 2025-05-22
Vis: visualisation of graphs that aren't running yet


mwpb:
Uses:-

* Can click through from a Const node containing a graph.
* During the graph creation process.

PR #148, opened 2025-07-31
Feat: graphdata visualizer

philipp-seitz:
closes: #114 

Implements a read-only storage for graph data objects that can be handed to the visualizer.
Assumes a graph has a single output

### DIFF
--- a/tierkreis/tierkreis/controller/storage/graphdata.py
+++ b/tierkreis/tierkreis/controller/storage/graphdata.py
@@ -15,15 +15,6 @@ from tierkreis.controller.data.location import (
 from tierkreis.exceptions import TierkreisError
 from tierkreis.controller.storage.in_memory import NodeData
 
-# class NodeData(BaseModel):
-#     definition: NodeDef | None = None
-#     call_args: WorkerCallArgs | None = None
-#     is_done: bool = False
-#     has_error: bool = False
-#     metadata: dict[str, Any] = Field(default_factory=dict)
-#     error_logs: str = ""
-#     outputs: dict[PortID, bytes | None] = Field(default_factory=dict)
-
 
 class GraphDataStorage:
     def __init__(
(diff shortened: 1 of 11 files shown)


tierkreis/tierkreis/controller/storage/graphdata.py
```diff
@@ -0,0 +1,237 @@
...
+    def write_worker_call_args(
+        self,
+        node_location: Loc,
+        function_name: str,
+        inputs: dict[PortID, OutputLoc],
+        output_list: list[PortID],
+    ) -> Path:
+        raise NotImplementedError("GraphDataStorage is read only storage.")
+
+    def read_worker_call_args(self, node_location: Loc) -> WorkerCallArgs:
```
mwpb, line 51: Just as a quick check more than anything: is there a situation where this gets successfully called? My assumption is that none of the function nodes get started.

tierkreis/tierkreis/controller/storage/graphdata.py
```diff
@@ -0,0 +1,237 @@
...
+                        outputs={"body": graph.model_dump_json().encode()},
+                    )
+                case "const" | "function" | "input" | "output" | "ifelse" | "eifelse":
+                    pass
+                case _:
+                    assert_never(node)
+
+        # Have found the node, now populate the node data correctly
+        if node_location not in self.nodes:
+            outputs: dict[PortID, bytes | None] = {
```
mwpb, line 202: This looks very similar to `_build_outputs`, can it (and the `*` check) be moved into there?

mwpb (commented):
Thanks for separating this out Philipp. It is looking much clearer! Have left a fair few comments, some of which are exploratory to try to figure out if we can simplify in certain places. In particular, `_node_from_loc` is much simpler than it looks now. In particular removing the caching makes things quite short. (The business with the MAP node string labels is the next obvious simplification we can make but that is outside the scope of this PR.)

tierkreis/tierkreis/controller/storage/graphdata.py
philipp-seitz: Yes, I missed this one. Although the tierkreis error is necessary, as this is caught by the visualizer

mwpb (approved):

PR #152, opened 2025-08-04
docs: python worker library

mwpb:


docs/source/tutorial/map.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Generating stubs
+
+Since this worker uses the Tierkreis Python library, we can automatically generate stub files using the following command.
+The stub files will provide us with type hints in the graph building process later on.
+
+```{code-cell}
+# Generate stubs file.
+!cd ../../../examples/example_workers/auth_worker && VIRTUAL_ENV="" uv run main.py --stubs-path ../../../docs/source/tutorial/auth_stubs.py
```
johnchildren, line 29: I guess the VIRTUAL_ENV="" is to avoid overriding the workspace virtual environment? That seems like a bit of an unfortunate detail. I wonder if we could extend the CLI tool to manage this process a bit, including for the formatting and linting?

examples/example_workers/hello_world_worker/main.py
```diff
@@ -19,5 +19,5 @@ def greet(greeting: str, subject: str) -> str:
     return greeting + subject
 
 
-if __name__ == "__main__":
+if __name__ == "__main__" and "__file__" in globals():
```
johnchildren, line 22: what does this new condition do? I've not seen it before.

docs/source/worker/hello_world.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+We use the `Worker.app` method to turn our Python programme into a legitimate Tierkreis worker.
+
+```{code-cell} ipython3
+if __name__ == "__main__" and "__file__" in globals():
+    worker.app(argv)
+```
+
+(Usually we don't need to check the `globals` object but in this case we want to make sure `.app` doesn't run in an interactive notebook.)
```
johnchildren, line 48: Ah I guess this is the answer to my earlier question.

docs/source/tutorial/map.md
mwpb: Good idea. I'll give it a go. The VIRTUAL_ENV is just to silence a warning from `uv` when it sees a mismatch between the Tierkreis venv and the worker venv.

tierkreis/tierkreis/worker/worker.py
```diff
@@ -121,6 +123,13 @@ def write_stubs(self, stubs_path: Path) -> None:
         with open(stubs_path, "w+") as fh:
             fh.write(format_namespace(self.namespace))
 
+        ruff_binary = shutil.which("ruff")
```
johnchildren, line 126: https://github.com/astral-sh/ruff/blob/94947cbf65388988cbe926302831bc075547acd8/python/ruff/__main__.py#L8

Interestingly it seems like ruff does provide a method for finding ruff.

johnchildren (approved):

PR #167, opened 2025-08-15
Feat: hpc executor

philipp-seitz:
This is a new draft for a template file-based HPC executor.
The idea is still pretty rough and inspired by [PSI/J](https://github.com/ExaWorks/psij-python), although it doesn't support pjsub.

There are some open questions still:
- Should we consider containerized environments like Apptainer?
- Should we, by default, assume MPI is enable and make it part of the job spec?
- Do we consider other resources like GPUs
- Do we want a batch submission -> if so, this would probably require some other worker

### DIFF
--- a/tierkreis/tierkreis/controller/executor/hpc/pbs.py
+++ b/tierkreis/tierkreis/controller/executor/hpc/pbs.py
@@ -1,6 +1,7 @@
 from pathlib import Path
-from typing import Callable
-from tierkreis.controller.executor.hpc.hpc_executor import run_hpc_executor
+
+# from typing import Callable
+# from tierkreis.controller.executor.hpc.hpc_executor import run_hpc_executor
 from tierkreis.controller.executor.hpc.job_spec import JobSpec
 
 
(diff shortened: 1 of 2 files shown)


tierkreis/pyproject.toml
```diff
@@ -14,6 +14,7 @@ authors = [
 
 requires-python = ">=3.10,<3.13"
 dependencies = [
+    "jinja2>=3.1.6",
```
mwpb, line 17: I'd avoid a dependency on jinja here. We aren't dealing with untrusted input and don't need HTML/JS syntax highlighting etc..  so we may as well generate the files using Python .

tierkreis/tierkreis/controller/executor/hpc/adapter.py
```diff
@@ -0,0 +1,23 @@
...
+    pjsub_large_spec,
+    pjsub_small_spec,
+)
+from tierkreis.controller.executor.hpc.protocol import TemplateAdapter
+
+_template_dir = "./tierkreis/tierkreis/controller/executor/hpc"
+
+PJSUB_ADAPTER = TemplateAdapter("pjsub.jinja", "pjsub", template_dir=_template_dir)
+PBS_ADAPTER = TemplateAdapter("pbs.jinja", "qsub", template_dir=_template_dir)
+SLURM_ADAPTER = TemplateAdapter("slurm.jinja", "sbatch", template_dir=_template_dir)
```
mwpb, line 14: These other adapters are nice... but before merging to main we will have to test them all of course...

tierkreis/tierkreis/controller/executor/hpc/hpc_executor.py
```diff
@@ -0,0 +1,71 @@
...
+            datefmt="%Y-%m-%dT%H:%M:%S%z",
+            filename=self.logs_path,
+            filemode="a",
+            level=logging.INFO,
+        )
+        logger.info("START %s %s", launcher_name, node_definition_path)
+
+        self.spec.command += str(node_definition_path)
+
+        with NamedTemporaryFile(
```
mwpb, line 41: When I was using Fugaku, I had trouble using the standard temporary directories (such as `/tmp`) because the cleaning was a bit too aggressive. That was for the workflow storage though, which is a bit different as those files need to be present for longer. Maybe we should be cautious in this case also?

tierkreis/tierkreis/controller/executor/hpc/job_spec.py
```diff
@@ -0,0 +1,64 @@
+from pathlib import Path
+import platform
+from pydantic import BaseModel, Field
+
+
+class JobSpec(BaseModel):
```
mwpb, line 6: Is this one spec for all HPC executors? If so that I'd suggest a less 'risky' approach and create separate specs for each type of HPC executor we are going to support.

tierkreis/tierkreis/controller/executor/hpc/job_spec.py
```diff
@@ -0,0 +1,64 @@
...
+            "--mpi": "max-proc-per-node=4",
+            "-z": "jid",
+            "--llio": "sharedtmp-size=64Gi",
+        },
+    )
+
+
+def pjsub_small_spec() -> JobSpec:
+    arch = platform.machine()
+    uv_path = Path.home() / ".local" / f"bin_{arch}" / "uv"
```
mwpb, line 47: Okay so this one encourages the user to use `uv`. That's okay but we will need others also, probably one that just executes some Bash.

mwpb (commented):
Thanks Philipp, just looking over this draft I've a few small comments. My main high-level comment would be to initially focus on one of the HPC executors (pjsub/Fugaku would be the natural choice). I suspect that each HPC setup will have its own idiosyncrasies and will probably merit custom code in each case.

tierkreis/tierkreis/controller/executor/hpc/hpc_executor.py
philipp-seitz: I set the `dir=launcher_path` for that reason, from the docs:
>  the file will be created in that directory;
so this should always exist, and as soon as the job has been submitted, it's no longer important (and deleted)

mwpb (commented):
Some (very) minor comments that were easier to do here than on Slack.

PR #198, opened 2025-09-16
fix: environment variables

mwpb:
Lower case x is required when setting the env vars for Spack at least.


mwpb:
Multiple env vars appears to work.

mwpb:
(As an aside, I think that there is still a case to be made for having the slurm, pbs and pjsub executors largely separate. Indeed a large part of the motivation of having them separate is that we don't need to keep all the other executors in our minds when making changes to a single one. Furthermore we can hedge our bets by making a separate psi/j executor also and if that becomes a standard we can just remove the others.)

PR #200, opened 2025-09-16
WIP: Add hot reloading command to the CLI

johnchildren:


tierkreis_visualization/tierkreis_visualization/cli.py
```diff
@@ -1,8 +1,15 @@
...
+from watchfiles import run_process
 
-from tierkreis_visualization.main import start
+from tierkreis.cli.tkr import load_graph
+from tierkreis_visualization.main import start, visualize_graph
+
+
+def _watch_graph(graph_location: str) -> None:
+    graph = load_graph(graph_location)
+    visualize_graph(graph.get_data())
```
philipp-seitz, line 12: Does this work correctly? From the type hints, `load_graph` returns GraphData, which does not have a `get_data()` method.

philipp-seitz (changes requested):

tierkreis_visualization/tierkreis_visualization/cli.py
johnchildren: It depends on whether the object you import is graphdata or graphbuilder, but I've been testing it with graphbuilder graphs and it works fine.

We should try to make sure that graphbuilder is always the public API though and clear up this confusion.

philipp-seitz: I assumed as much, I tried with GraphData and it failed for me

PR #205, opened 2025-09-19
chore: Fix direnv environments

johnchildren:
Fixes the issue where environments would occasionally break when opening files in different projects.


mwpb (approved):
(Looks reasonable. However, I don't use devenv so cannot really test.)

PR #221, opened 2025-09-29
fix: ci steps not found

mwpb:
Add explicit steps again. Will spend some time making this reusable later today.


.github/workflows/docs.yml
```diff
@@ -2,6 +2,7 @@ on:
   release:
     types: [created]
   workflow_dispatch:
+  push:
```
johnchildren, line 5: do you still need this `push` ?

johnchildren (approved):

PR #236, opened 2025-10-16
feat: use pytket worker for simulation

mwpb:
Move simulation to the `pytket` worker, which brings in Qulacs simulation 'for free'. 

* Rearrange documentation to explain that simulation can be done with the `pytket` worker.
* Parameterise simulation example using the `BackendConfig`. Also parameterise prepackaged simulation graph.
* (Improve error handling: fix bug in union types and be more verbose when the graph has errored.)
* Create wrappers around pytket backend methods. Indicate that they are only intended for use with the simulators Aer and Qulacs for now. Essentially -- for the more complicated backends -- we'd like to reuse code from nexus-pytket for instantiating them. Add wrappers to worker.

### DIFF
--- a/tierkreis/tierkreis/graphs/simulate/compile_run.py
+++ b/tierkreis/tierkreis/graphs/simulate/compile_run.py
@@ -11,23 +11,21 @@ type BackendResult = OpaqueType["pytket.backends.backendresult.BackendResult"]
 type Circuit = OpaqueType["pytket._tket.circuit.Circuit"]
 
 
-class AerJobInputs(NamedTuple):
+class SimulateJobInputs(NamedTuple):
     circuits: TKR[list[Circuit]]
     n_shots: TKR[list[int]]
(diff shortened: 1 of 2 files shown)


examples/simulate_in_parallel.py
```diff
@@ -1,21 +1,20 @@
...
 from tierkreis.executor import UvExecutor
 
-config = AerConfig()
 circuit = circuit_from_qasm(Path(__file__).parent / "data" / "ghz_state_n23.qasm")
 circuits = [circuit] * 10
 
 
-def main():
-    g = aer_compile_run()
+def main(config: AerConfig | QulacsConfig):
```
johnchildren: I think calling this "main" is perhaps a bit unclear, I would not expect "main" to have any arguments, perhaps it is worth re-naming this function to "simulate_example" and have different configs passed in a different "main" function?

PR #248, opened 2025-11-13
Refactor and get hot reloading working

mwpb:
A quick fix for hot reloading a graph under construction as well as a bit of clean up of how we load the Python app.

* Add an extra output to `examples/defaults.py` for an easy example of hot reloading a graph under construction.
* (Use vite to build into the directory served by FastAPI rather than spin up its own server. This removes the need for weakening CORS as well as giving us a higher fidelity testing setup.)
* Do not set up web socket when we are using the graph data visualiser.
* Separate FastAPI app config, FastAPI app shell and uvicorn shell. Define two FastAPI apps for file storage and graph data.
* Move storage wrapper to separate file.
* Add CLI commands `tkr_viz_dev` and `tkr_viz_graph` to the `pyproject.toml`. The command `tkr_viz_graph` requires a CLI argument that specifies the Python file and variable referring to the graph under construction.
* (Also fix client-side routing.)

### DIFF
--- a/tierkreis_visualization/frontend/vite.config.ts
+++ b/tierkreis_visualization/frontend/vite.config.ts
@@ -1,15 +1,18 @@
-import path from "path"
-import tailwindcss from "@tailwindcss/vite"
-import { defineConfig } from 'vite'
-import react from '@vitejs/plugin-react'
+import path from "path";
+import tailwindcss from "@tailwindcss/vite";
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
 
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react(), tailwindcss()],
+  build: {
+    outDir: "../tierkreis_visualization/static/dist",
+  },
   resolve: {
     alias: {
       "@": path.resolve(__dirname, "./src"),
     },
   },
-  base: '',
-})
+  base: "",
+});


tierkreis_visualization/tierkreis_visualization/app.py
```diff
@@ -0,0 +1,45 @@
...
+    if graph_specifier is None:
+        return app
+
+    app.state.get_storage_fn = graph_data_storage_fn(graph_specifier)[0]
+    app.state.storage_type = StorageType.GRAPHDATA
+    return app
+
+
+app = get_filestorage_app()
+app_graph_data = get_graph_data_app()
```
johnchildren: These are both instances of `app` that exist in memory regardless of which one you end up using?

tierkreis_visualization/pyproject.toml
```diff
@@ -21,4 +21,5 @@ tierkreis = { workspace = true }
 
 [project.scripts]
 tkr-vis = "tierkreis_visualization.main:start"
-
+tkr_viz_dev = "tierkreis_visualization.main:dev"
```
johnchildren: `tkr-viz` versus `tkr_viz` might be a bit confusing. Though I think possibly these scripts might make more sense as subcommands of the main `tkr` CLI anyway in future?

tierkreis_visualization/tierkreis_visualization/app.py
mwpb: Hmm... perhaps worth optimising this. Although one has to use the string based specifier in `uvicorn.run` when we want `reaload=True`, it turns out that one can pass a factory function in, which neatly gets around this problem.

PR #256, opened 2025-11-19
feat: first step for ndarray values

mwpb:
Add initial support for NumPy's `ndarray` using a protocol to avoid a library dependency. The next step (in a different PR) will be to allow customisation of the (de)serialiser via `Annotation`s.

* Document how the `ndarray` type behaves with respect to serialisation and code generation.
* Create example SciPy worker and example graph that passes `ndarray`s through various levels of nesting.
* Create `NdarraySurrogate` protocol. We
  * allow `ndarray`s as a PType
  * serialise using `ndarray.dumps`
  * deserialise using `pickle.loads`
* Fix deserialisation of `Struct`, which we recently allowed to be nested.

### DIFF
--- a/examples/scipy_graph.py
+++ b/examples/scipy_graph.py
@@ -2,7 +2,6 @@ from pathlib import Path
 from typing import NamedTuple
 from uuid import UUID
 from tierkreis.builder import GraphBuilder
-from tierkreis.consts import PACKAGE_PATH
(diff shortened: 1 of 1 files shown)


mwpb:
No important reason, just pre-emptively trying to reduce the number of example workers we have... Could quite easily change to just a numpy worker.

PR #260, opened 2025-12-04
feat: extra visualiser pages and nested updating

mwpb:
Many small visualiser improvements. Some new features also. For instance: a sortable workflows list page and a bit better live updating of nested graphs. Various further improvements possible of course (especially any styling that I've introduced!) but I think the following is more than enough for one PR:-

* (Fix `tsc` errors.)
* A double click on eval/loop/map nodes opens the node at the top level. (Existing expand behaviour maintained using the plus/minus buttons.) Use breadcrumbs to aid navigation.
* Generate an openapi spec when the FastAPI dev server is running. Generate TS types from the openapi spec when the Vite dev build is running. Use these types in the `openapi-fetch` library to type check tanstack query and various one-off fetch requests.
* Move to tanstack router (file path based) for type checking of client side routing. 
* Use `node_location` as the node ID to describe nested structures. Pass `node_type` from backend to frontend.
* Move to a declarative approach to expanded nodes. The web socket now just invalidates. Enables simpler live updating of nested structures and back button navigation has finer granularity.
* The logic in `graph/updateGraph.ts` has been kept elementary for now rather than choosing a particular abstraction. I tried a few different ways that resulted in 'better code' but were harder to understand from the perspective of graph-theory semantics.
* The `/` route now serves a table of all the workflows, which is sortable by name, ID and timestamp. (Default sorting is timestamp.) The last column is a list of links to errored nodes, which should speed up graph debugging.
* Remove all templating code. Remove old JS and CSS.
* Work around server hanging waiting for WebSocket connections to close. In dev we just always send SIGKILL. In 'prod' we wait 10 seconds before SIGKILL. More thinking required for proper solution.

### DIFF
--- a/tierkreis_visualization/frontend/src/routes/workflows/_.$wid.nodes.$loc/-components/GraphView.tsx
+++ b/tierkreis_visualization/frontend/src/routes/workflows/_.$wid.nodes.$loc/-components/GraphView.tsx
@@ -82,6 +82,7 @@ export const GraphView = (props: {
         onNodesChange={props.onNodesChange}
         onNodeDrag={onNodeDrag}
         minZoom={0.01}
+        defaultEdgeOptions={{zIndex: -2}}
         fitView
       >
         <Background />


tierkreis_visualization/frontend/src/routes/workflows/_.$wid.nodes.$loc/-components/NodePage.tsx
```diff
@@ -0,0 +1,93 @@
...
+    workflow_id + node_location_str,
+    { defaultValue: { nodes: [], edges: [] } }
+  );
+
+  const onNodesChange = useCallback((changes: NodeChange<BackendNode>[]) => {
+    setG((gSnapshot: Graph) => {
+      const ns = applyNodeChanges(changes, gSnapshot.nodes);
+      return { nodes: ns, edges: gSnapshot.edges };
+    });
+  }, []);
```
aidanCQ:
```suggestion
  }, [setG]);
```

tierkreis_visualization/frontend/src/data/api.ts
```diff
@@ -0,0 +1,51 @@
...
+  const params = { path: { workflow_id, node_location_str } };
+  const res = await fetchClient.GET(
+    "/api/workflows/{workflow_id}/nodes/{node_location_str}",
+    { params }
+  );
+  return res.data ?? { nodes: [], edges: [] };
+};
+
+export const listWorkflowsQuery = () => $api.useQuery("get", "/api/workflows/");
+export const logsQuery = (workflow_id: string) =>
```
aidanCQ: This is just convention btw but in general any function abstracting a hook also uses the `use` prefix to indicate this.
```suggestion
export const useLogsQuery = (workflow_id: string) =>
```

tierkreis_visualization/tierkreis_visualization/data/models.py
```diff
@@ -1,23 +1,25 @@
...
 
 NodeStatus = Literal["Not started", "Started", "Error", "Finished"]
 
 
 class PyNode(BaseModel):
-    id: int
+    id: Loc
     status: NodeStatus
     function_name: str
+    node_type: str
```
aidanCQ: Similar to NodeStatus you could literally type the union and remove casting in the frontend for a stricter contract.
```suggestion
    node_type: Literal["eval", "loop", "map"]
```

tierkreis_visualization/frontend/src/graph/updateGraph.ts
```diff
@@ -0,0 +1,131 @@
+import { bottomUpLayout } from "./layoutGraph";
+import { Graph } from "./models";
+import { loc_children, loc_peek } from "@/data/loc";
+import { PyEdge, PyNode } from "@/data/api_types";
+import { getContainingNodes } from "@/nodes/layout";
+
+const DELETE_TAG = "fordeletion";
+
+export const amalgamateGraphData = (
```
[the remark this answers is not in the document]
mwpb: Discussed in person: will be in a follow up PR.

aidanCQ:
I've noticed there are two lock files, is this intentional?

Screenshot 2026-01-05 at 10 42 40

justfile
```diff
@@ -33,8 +33,8 @@ serve:
 
 [working-directory:'tierkreis_visualization/frontend']
 prod:
-  bun install
-  bunx vite build .
+	npm i
+	npm run build
```
philipp-seitz: to be consistent with the devenv add npm there and remove bun (although I would prefer bun)
```
    npm.enable = true;
```

tierkreis_visualization/frontend/src/graph/layoutGraph.ts
```diff
@@ -99,25 +101,59 @@ function resizeGroupNodesToFitChildren(
...
+  edges.forEach((edge) => {
+    if (
+      nodeIds.has(edge.source) &&
+      nodeIds.has(edge.target) &&
+      !loc_peek(edge.source)?.includes("L")
+    ) {
+      dagreGraph.setEdge(edge.source, edge.target);
+    }
+  });
+  dagre.layout(dagreGraph, { disableOptimalOrderHeuristic: true });
```
philipp-seitz: what was the reason for this again? I think I get better layouts without setting this option

untitled.txt
philipp-seitz: I assume this was added on accident?

philipp-seitz (commented):
Also a bugI discovered: when expanding nested nodes, for some reason the edge labels get put behind the edges. But I can't find a simple reason why this happens (also I'm not sure it did before)

philipp-seitz (approved):

Issue #201, opened 2025-09-17
Error Logs are not consistent


philipp-seitz:
In some instances, workers/executors write to `_error` (the storage error path) and not to `errors` (the storage error_logs path).

Issue #181, opened 2025-09-05
Expose Logging/Error to workers


philipp-seitz:
Currently workers running in different locations (e.g. compute nodes) cannot access the shared logs

Issue #62, opened 2025-05-08
Clean up loggers in controller


philipp-seitz:
Sometimes it is unclear where the output will end up. E.g. most of it ends up in the storage’s log_path.

mwpb:
Idea for worker logging: use the executors to:-
* pipe STDOUT to the global log file
* pipe STDERR to the global log file and also the `errors` file in the node directory.

and remove `basicConfig` from `tierkreis/tierkreis/worker/worker.py`.

Then instruct worker authors use STDOUT/STDERR as appropriate.

PR #261, opened 2025-12-12
chore: error handling

philipp-seitz:
This cleans up the calls to `logging.basicConfig`, removes some unused logging calls and redirects output from stderr into the central workflow log.
The controller will now also raise an error when the graph execution stops.

closes: #230, closes: #62 , closes: #201, closes: #181

To be discussed:
- ~Currently stderr get's writen to `loc/errors` while the direct crashing reason is written to `loc/_error`. Should we keep it this way?~

### DIFF
--- a/docs/source/logging_and_errors.md
+++ b/docs/source/logging_and_errors.md
@@ -5,25 +5,10 @@
 Tierkreis stores a central log of the execution of a graph.
 The log is located at `<checkpoint_dir>/<workflow_id>/logs`.
 By default, the log level is set to `INFO`.
-To change the log level you can get the tierkreis logger:
-
-```python
-import logging
-
-logger = logging.getLogger("tierkreis")
-logger.setLevel(...)
-```
 
 ### Workers
 
-Similarly in a worker you can access the logger or overwrite it with your own.
-
-```python
-worker.logger.setLevel(...)
-worker.setLogger(...)
-```
-
-When running a python worker, they will check the environment variable `$TKR_LOG_LEVEL`, `$TKR_LOG_FORMAT` and `$TKR_DATE_FORMAT` and add a streaming handler to the root logger.
+When running a python worker, Tierkreis will check the environment variables `$TKR_LOG_LEVEL`, `$TKR_LOG_FORMAT` and `$TKR_DATE_FORMAT` and add a streaming handler to the root logger with the appropriate level and formatter.
 You can either set this manually or provided it as part of the `env` argument of an executor.
 
 ## Error Handling
@@ -111,7 +96,7 @@ logging.basicConfig(
 )
 ```
 
-For example by uncommenting lines 24-28 in the example.
+For example by uncommenting lines 24-28 in the example and adding `import logging`.
 
 #### Resume a Workflow
 


docs/source/worker/external_workers.md
```diff
@@ -70,10 +70,19 @@ There are two possible scenarios to parse input and outputs for workers.
...
+Tierkreis typically hands over a single argument to the invoked script: the [`WorkerCallArgs`][#tierkreis.controller.data.location.WorkerCallArgs]
 This contains all the information specified in the the contract of a worker.
 Parsing this file will provide the locations of inputs and expected outputs.
 
+### Errors and Logging
+
+Part of the specification for `WorkerCallArgs` are path descriptions for `logs_path` and `errors_path`.
+These point to files that the Tierkreis controller will collect logs and errors in.
+To make worker errors visible, you can use those paths to write the information.
+Alternatively, the controller will redirect any output to `stdout` to `logs_path` and `stderr` to `errors_path` automatically.
```
mwpb: Note for tomorrow: we'd like both `stdout` and `stderr` to go to `logs_path`. Possibly only `stderr` to go to `errors_path`, though we may like to think if we want both here also.

tierkreis/tierkreis/controller/executor/uv_executor.py
```diff
@@ -30,15 +31,10 @@ def run(
...
-            datefmt="%Y-%m-%dT%H:%M:%S%z",
-            filename=self.logs_path,
-            filemode="a",
-            level=logging.INFO,
-        )
         self.errors_path = (
-            self.logs_path.parent.parent / worker_call_args_path.parent / "errors"
+            self.logs_path.parent.parent
+            / worker_call_args_path.parent
+            / "logs"  # made we should change this
```
mwpb: Good change. Though it does entail changing a lot of the documentation too. We should probably start referring to 'global logs' vs 'local/task logs' rather than logs vs errors.

docs/source/worker/hello_world.md
```diff
@@ -53,6 +53,14 @@ The complete worker file is as follows:
 :language: python
 ```
 
+### Logging and Errors
+
+The Tierkreis controller will automatically collect logs and errors by redirecting any output to `stdout` to `logs_path` and `stderr` to `errors_path` automatically.
```
mwpb: Along the same lines, I think this needs updating also.

docs/source/logging_and_errors.md
```diff
@@ -0,0 +1,112 @@
+# Logging and Errors
+
+## Logging
+
+Tierkreis stores a central log of the execution of a graph.
+The log is located at `<checkpoint_dir>/<workflow_id>/logs`.
+By default, the log level is set to `INFO`.
+To change the log level you can get the tierkreis logger:
```
mwpb: An additional feature that might be useful is for this logger to respect `TKR_LOG_LEVEL` also.

docs/source/logging_and_errors.md
```diff
@@ -0,0 +1,128 @@
...
+### Workers
+
+Similarly in a worker you can access the logger or overwrite it with your own.
+
+```python
+worker.logger.setLevel(...)
+worker.setLogger(...)
+```
+
+When running a python worker, they will check the environment variable `$TKR_LOG_LEVEL`, `$TKR_LOG_FORMAT` and `$TKR_DATE_FORMAT` and add a streaming handler to the root logger.
```
mwpb:
```suggestion
When running a python worker, Tierkreis will check the environment variables `$TKR_LOG_LEVEL`, `$TKR_LOG_FORMAT` and `$TKR_DATE_FORMAT` and add a streaming handler to the root logger with the appropriate level and formatter.
```

mwpb (approved):
Thanks Philipp, looks good!

PR #266, opened 2026-01-02
feat: Rust internals

johnchildren:
<description pending>


tierkreis_core/rust/identifiers.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        pub fn __str__(&self) -> String {
+            self.0.to_string()
+        }
+    }
+
+    #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, FromPyObject, IntoPyObject)]
+    pub enum ExteriorOrValueRef {
+        Exterior(ExteriorRef),
```
johnchildren: I would like to remove `Exterior` as an option potentially as it would simplify some of the code a bit, this requires some changes to the controller code itself however.

tierkreis_core/rust/location.rs
```diff
@@ -0,0 +1,345 @@
...
+            let parts = k.split_terminator(".");
+            let mut steps = VecDeque::new();
+            for part in parts {
+                steps.push_back(NodeStep::new(part)?);
+            }
+            Ok(Self { steps })
+        }
+
+        #[allow(non_snake_case)]
+        pub fn N(&self, raw_index: i64) -> PyResult<Self> {
```
johnchildren: If possible I would like to remove these methods in favor of the `append_node` / `append_loop` / ... methods. I think it might only take a little bit more work to do so.

tierkreis/tests/controller/test_graphdata.py
```diff
@@ -1,10 +1,9 @@
 import pytest
-from tierkreis.exceptions import TierkreisError
-from tierkreis.controller.data.graph import GraphData
+from tierkreis_core import GraphData
 
 
 def test_only_one_output():
-    with pytest.raises(TierkreisError):
+    with pytest.raises(ValueError):
```
johnchildren: Philosophical question here about whether we should use the standard python errors or custom ones. I can raise a custom exception in the rust code instead but this was easier for me to do at least initially.

tierkreis/tests/controller/test_graphdata_storage.py
```diff
@@ -2,32 +2,25 @@
...
-        ("-.N4.M0", simple_map(), Eval((-1, "body"), {})),
-        ("-.N4.M0.N-1", simple_map(), Eval((-1, "body"), {})),
-    ],
-)
-def test_read_nodedef(node_location_str: str, graph: GraphData, target: str) -> None:
-    loc = Loc(node_location_str)
-    storage = GraphDataStorage(UUID(int=0), graph)
-    node_def = storage.read_node_def(loc)
-    assert node_def == target
+# @pytest.mark.parametrize(
```
johnchildren: These tests need to be re-enabled, I don't have an easy way to construct nodes outside of a graph (other than for eval nodes) currently however.

tierkreis/tierkreis/controller/storage/adjacency.py
```diff
@@ -1,46 +1,19 @@
...
-
-
 def unfinished_inputs(
     storage: ControllerStorage, loc: Loc, node: NodeDef
 ) -> list[ValueRef]:
-    ins = in_edges(node).values()
-    ins = [x for x in ins if x[0] >= 0]  # inputs at -1 already finished
-    return [x for x in ins if not storage.is_node_finished(loc.N(x[0]))]
+    ins = node.in_edges.values()
+    ins = [x for x in ins if isinstance(x, ValueRef)]  # Only look an Values on Edges
```
johnchildren: I think this comment is both incorrect and needs expansion

tierkreis/tests/controller/test_graphdata_storage.py
```diff
@@ -2,32 +2,34 @@
...
+    NodeDescription,
+    ValueRef,
+    NodeIndex,
+)
 
 
 @pytest.mark.parametrize(
     ["node_location_str", "graph", "target"],
     [
-        ("-.N0", simple_eval(), Const(0, outputs=set(["value"]))),
```
johnchildren: I guess I should test the full definition as this was also looking at outputs before

tierkreis/tests/controller/test_locs.py
```diff
@@ -1,27 +1,26 @@
...
 
 node_location_1 = Loc()
-node_location_1 = node_location_1.N(1)
-node_location_1 = node_location_1.L(0)
-node_location_1 = node_location_1.N(3)
-node_location_1 = node_location_1.L(2)
-node_location_1 = node_location_1.N(0)
-node_location_1 = node_location_1.M(7)
-node_location_1 = node_location_1.N(10)
+node_location_1 = node_location_1.N((1))
```
johnchildren: extra brackets are accidental here

tierkreis_core/rust/location.rs
```diff
@@ -0,0 +1,370 @@
...
+            _handler: &Bound<'_, PyAny>,
+        ) -> PyResult<Py<PyAny>> {
+            let pydantic = PyModule::import(cls.py(), "pydantic_core")?;
+            let core_schema = pydantic.getattr("core_schema")?;
+
+            let validate_fn = cls.getattr("_validate")?;
+            // Used to validate the output of the schema
+            let any_schema = core_schema.call_method0("any_schema")?;
+
+            let serialization = core_schema.call_method0("to_string_ser_schema")?;
```
johnchildren: I probably could add an example to the pydantic schema here if we want Loc's to appear properly in the API docs. A bit of a nice to have though.

tierkreis_visualization/tierkreis_visualization/data/loop.py
```diff
@@ -47,15 +47,16 @@ def get_loop_node(
             id=new_location,
             status=last_status,
             function_name=f"L{i}",
-            node_location=new_location,
+            node_location=str(new_location),
```
johnchildren: possibly unnecessary str call

tierkreis_visualization/tierkreis_visualization/data/map.py
```diff
@@ -14,26 +13,26 @@ class MapNodeData(BaseModel):
...
+    map_eles = outputs_iter(storage, parent.extend_from_ref(first_ref))
+    description = storage.read_node_description(loc.M(0))
     nodes: list[PyNode] = []
     for i, ele in map_eles:
         node = PyNode(
             id=loc.M(i),
             status="Started",
             function_name=ele,
-            node_location=loc.M(i),
+            node_location=str(loc.M(i)),
```
johnchildren: possibly unnecessary str call

tierkreis/tests/controller/test_eagerifelse.py
```diff
@@ -58,7 +50,7 @@ def test_eagerifelse_long_running(input: dict[str, PType], output: int) -> None:
 
 def test_eagerifelse_nodes() -> None:
     g = simple_eagerifelse()
-    storage = ControllerFileStorage(UUID(int=150), name="simple_if_else")
+    storage = ControllerFileStorage(UUID(int=151), name="simple_if_else")
```
acl-cqc: nit: should the name (`"simple_if_else"`) here reflect the name of the test (`eagerifelse`)? and similarly in `test_ifelse_nodes` below

tierkreis/tests/controller/test_graphdata_storage.py
acl-cqc: functions? e.g. first instantiation `("-.N0", simple_eval, lambda ...: ...Const(0, outputs=set(["value"]))...)`

tierkreis/tests/controller/test_locs.py
acl-cqc: in rust the linter would remove them for you!!

tierkreis/tierkreis/builder.py
```diff
@@ -33,7 +34,7 @@ def out() -> type[Out]: ...
 
 @dataclass
 class TypedGraphRef[Ins: TModel, Outs: TModel]:
-    graph_ref: ValueRef
+    graph_ref: ExteriorRef | ValueRef
```
acl-cqc: So AFAICS ExteriorRef is a reference to an input to the graph - so an outport of the Input node. I guess that means we haven't assigned the Input node an index yet - maybe we should (require users to) do so earlier / more aggressively? (Or do you not have Input nodes at all, hmmmm)

tierkreis/tierkreis/controller/storage/adjacency.py
```diff
@@ -1,48 +1,24 @@
...
 
 
-def outputs_iter(storage: ControllerStorage, loc: Loc) -> list[tuple[int, PortID]]:
+# TODO: gotta be a better way to do this?
+def outputs_iter(
+    storage: ControllerStorage, loc: Loc
+) -> list[tuple[NodeIndex, PortID]]:
     eles = storage.read_output_ports(loc)
-    return [(int(x.split("-")[-1]), x) for x in eles]
+    return [(NodeIndex(int(x.split("-")[-1])), x) for x in eles]
```
acl-cqc: AFAICS `read_outport_ports` returns a list of `PortId` which I would not expect to contain `NodeIndex` at all so it seems odd to construct the latter from one of the former! I guess it mirrors the old code so is no more (or less) suprrising and that seems like evidence that it's correct....

(Unless PortID = Node and Port, which in Hugr we called NodePort. In that case...the `[-1]` suggests a sequence of `-`-separated nodes, i.e. a Location; you could give PortID a method to get the Location and then have a method on Location to get the last node from that? But is that really what you're doing? Do you mean `[(NodeIndex(int(loc.split("-")[-1])), x) for x in eles]` which is quite different (and the NodeIndex is loop invariant)?

tierkreis_core/rust/graph.rs
```diff
@@ -0,0 +1,734 @@
...
+            if_true: ValueRef,
+            if_false: ValueRef,
+        },
+        EagerIfElse {
+            pred: ValueRef,
+            if_true: ValueRef,
+            if_false: ValueRef,
+        },
+        Input {
+            name: PortID,
```
acl-cqc: The name is not conceptually a PortID at all, is it? It's just...a debug identifier or similar?

tierkreis_core/rust/graph.rs
```diff
@@ -0,0 +1,734 @@
...
+            .collect();
+        NodeDef::Eval {
+            body: ExteriorOrValueRef::Exterior(ExteriorRef("body".to_string())),
+            inputs: inputs_refs,
+        }
+    }
+
+    #[pyclass(eq, frozen)]
+    #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+    pub enum NodeDef {
```
acl-cqc: Ah I see. So these "Nodes" include their own input-edges; they are not nodes in the sense of "a graph is nodes plus edges".

Which gives some type safety, but generally means traversal is a bit harder - there's no representation of the output edges of a node except by looking *inside* a load of other Nodes. Not what we did in Hugr, but never mind....

I would say this is more a "task" or "task description" than a node, though. (You could have an enum of Node without the edges *as well*, if that's useful)

tierkreis_core/rust/graph.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    #[pymethods]
+    impl NodeDef {
+        /// Get the `inputs` attribute for a Node. This is different to
+        /// `in_edges` as it only considers the edges under the `inputs`
+        /// attribute and not things like the body of Eval nodes which
+        /// is considered a different kind of edge.
+        #[getter]
+        pub fn inputs(&self) -> IndexMap<PortID, ExteriorOrValueRef> {
```
acl-cqc: Yeah I have to wonder whether this concept of inputs vs body is even useful for many node types. I mean, obviously for say Func the two are the same, but it seems more accurate to say that a Func only has "in_edges" - AFAICS "inputs" are for some kind of node-specific internal use?

tierkreis_core/python/tierkreis_core/nodes.pyi
```diff
@@ -0,0 +1,135 @@
...
+            tierkreis_core.aliases.PortID,
+            identifiers.ExteriorRef | identifiers.ValueRef,
+        ],
+    ) -> None: ...
+    @property
+    def name(self, /) -> str: ...
+    @property
+    def inputs(
+        self, /
+    ) -> dict[
```
acl-cqc: is the `inputs` dict mutable? (does `foo.inputs[k]=v` affect other code that uses `foo.inputs`?) `Mapping` would be better if not...

tierkreis_core/rust/graph.rs
```diff
@@ -0,0 +1,734 @@
...
+                    };
+                    let (_, remaining) = remaining.pop_first()?;
+                    if remaining.steps().len() < 2 {
+                        return Ok(NodeDescription {
+                            definition: NodeDef::Eval {
+                                body: ExteriorOrValueRef::Exterior(ExteriorRef("body".to_string())),
+                                inputs: inputs.clone(),
+                            },
+                            outputs: outputs.clone(),
+                            outer_graph: Some(const_graph.clone()),
```
acl-cqc: Is that right? As per PR comment I think you should probably move this into python, but say I graph a graph G containing a const C which is a graph containing a const D which is also a graph....and I query with a Loc that identifies a node inside D...I think I'll get back graph C (not graph D, the one containing the node in question). Is that what you want?

acl-cqc (commented):
Thanks John, I can see you've done a lot of hard work here in particular introducing maturin etc. into what has been an all-python project. A few thoughts, mostly on the rust/python boundary and the new graph structure...
* As you say, quite a big PR. I wonder if you can split it in two, or indeed perhaps even remove half of it. Specifically, can you remove the porting of `Loc` into Rust? I think the new API is much nicer, but (a) you could implement that in python, as an orthogonal change; or (b) delay having a rust Loc until you really need it (i.e. some amount of controller code moves into Rust too).
   * The only *use* of `Loc` in Rust is for `query_node_description` which I reckon you could implement *in python* just querying the Rust graph structure "one level at a time". (Really I think the `Loc` is kind of a concept of *the controller*, not the graph; the latter needs to deal only with individual NodeId + PortId's)

* ExteriorRef. You hint at this in one of your own comments. I think in an actual graph, all NodeDefs would contain ValueRef's (not ExteriorRefs); what would it take to remove ExteriorRef?
   * Ah oh no, you allow a graph to refer *to itself* via `def ref()`, which is only valid because all the nodes *in which* the graph might be run have a graph input called `body`. Eeeek....can I say that again, eek? Ok - I think the thing to do here is to add a new input-less node, called say RecursiveGraphSelfReference (longer = more likely to alert the reader that something funny is going on....no ok, I jest, find a shorter name) that's a lot like Input i.e. produces the graph on its own output port. (Another advantage here is that the graph building does not rely on "all nodes that might run this graph must take it on an inport called body" - instead the actual graph is provided by the controller.)
   * In `new_eval_root` which creates a `NodeDef::Eval` containing a load of `ExteriorRefs`. So the thing here is that NodeDef is used both to represent a node in a graph (*with its edges* - not the usual idea of a graph being nodes + edges; really this is NodeAndInEdges rather than NodeRef), and also a description of a *task* (actual work to do). So (given the controller here is in python) I think a good thing to do would be to define a *python* struct for a task, which is basically a node that can be executed - has its inputs available - so actually this is just NodeRunData, but give it a map from PortId -> actual values/data (I think that's PortId -> (Loc,PortId)). There could be an easy constructor for NodeRunData from a NodeDef and a Loc which builds the input map by looking up the source, of each incoming edge in the NodeRef, in the *parent* Loc, as `start` currently does; and then `start` uses the input in the NodeRunData. (I'm keeping NodeDef as a field of NodeRunData even though the inputs/in-edges of the NodeDef will not be used, because it's not worth duplicating the enum). Then new_eval_root builds a *task* with the input map being *exterior* Loc+PortId
   ...then, ok, you need ExteriorRef, but only in python as part of NodeRunData, not in Rust or in the graph.

* As below, I think it'd be cleaner and have less risk of confusion if `NodeDef` didn't have both in_edges and inputs

tierkreis/tierkreis/controller/start.py
```diff
@@ -72,104 +76,115 @@ def start(
...
+            input_loc = parent.exterior()
+            storage.link_outputs(node_location, node.name, input_loc, node.name)
+            storage.mark_node_finished(node_location)
 
-    ins = {k: (parent.N(idx), p) for k, (idx, p) in node.inputs.items()}
+        case NodeDef.Output():
+            storage.mark_node_finished(node_location)
 
-    logger.debug(f"start {node_location} {node} {ins} {output_list}")
-    if node.type == "function":
```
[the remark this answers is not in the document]
johnchildren: I actually preferred the old style a bit, but I couldn't get it to work that way. The new style isn't exhaustive because of subclassing

tierkreis/tierkreis/controller/storage/walk.py
```diff
@@ -56,8 +55,12 @@ def walk_node(
         logger.error(f"\n\n{storage.read_errors(loc)}\n\n")
         return WalkResult([], [], [loc])
 
-    node = graph.nodes[idx]
-    node_run_data = NodeRunData(loc, node, list(node.outputs))
+    node = graph.get_nodedef(idx)
+    graph_outputs = (
+        set(outputs) if (outputs := graph.outputs(idx)) is not None else set()
+    )
+    # TODO: should we be assuming the are outputs to the graph at this point?
```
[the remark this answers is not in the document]
johnchildren: I think my brain must have melted mid comment, apologies, will clean up.

mwpb:
Thanks all. Whilst I recognise a lot of the concepts being talked about here (and agree with quite a few of the proposed improvements), I'm getting confused as multiple things are moving at once. Does the following list of potential improvements look roughly correct?

* Remove the -1 nodes. A few strategies here but one simple one might be: when starting an EVAL node, put the inputs into the relevant input nodes directly.
* Allow exterior references? I.e. to a path that isn't fully described by a `NodeIndex`.
* Consistently use one of `in_edges` vs `inputs`.
* Rework/remove recursion and self references.
* Rework the situation when the return type of the body of a MAP is a portmapping. Rework TList alongside.
* Improve the `Loc` datastructure. (+ port to Rust)
* Change graph representation. Maybe separate the representation (currently adjacency list) from the data required to run? (+ port to Rust)
* Output nodes of graphs. My guess is that we want to ensure that every graph always has one. Also I'd rather the author had to explicitly add it but don't have a very strong opinion about that.
* Remove the `fixed_inputs`.
* Implement a node for partial evaluation?

If so then how would we prefer to group this work?

tierkreis_core/rust/graph.rs
```diff
@@ -0,0 +1,715 @@
...
+                    ..
+                } => {
+                    let connected_nodedef = self.get_nodedef(value_ref.node_index())?;
+                    let const_graph = match connected_nodedef {
+                        NodeDef::Const {
+                            value: Value::Graph(graph),
+                        } => graph,
+                        _ => {
+                            return Err(PyValueError::new_err(
+                                "Const node connected to body port does not contain a graph",
```
acl-cqc: In this case it's not necessarily a Const node, right? (so any other node producing a graph, e.g. `if`, connected to an eval)

PR #274, opened 2026-01-07
feat: restart dependent nodes

mwpb:
Restart a node by: removing all nodes that fully depend on it and marking all nodes directly above it in the hierarchy as not finished.

* Create notebook to give an example of restarting a node in a graph.
* Add the outputs to the `NodeDef` so that we can recursively get the invalidated nodes. Add these recursive functions to the storage base class, which returns all the fully invalidated nodes.
* Implement a function on `Loc` to get all of the partially invalidated nodes. (Also add a few helper functions to `Loc`.)
* Create button on the info dialog for function nodes that restarts the given function node.
* Add endpoint to the vis server that restarts a node.

### DIFF
--- a/tierkreis/tierkreis/controller/storage/protocol.py
+++ b/tierkreis/tierkreis/controller/storage/protocol.py
@@ -42,7 +42,7 @@ class ControllerStorage(ABC):
     def delete(self, path: Path) -> None:
         """Delete the storage entry at the specified path.
 
-        Also delete any related data with keys of the form \"{path}/**/*\"."""
+        Also delete any related data of the form \"{path}/**/*\"."""
 
     @abstractmethod
     def exists(self, path: Path) -> bool:


tierkreis_visualization/frontend/src/nodes/eval.tsx
```diff
@@ -37,8 +37,7 @@ export function EvalNode({ data: node_data }: NodeProps<BackendNode>) {
           setHoveredId={node_data.setHoveredId}
         />
         <div className="grid justify-items-end">
-          <ZoomOutButton wid={wid} loc={loc} node_loc={node_loc} />
-        </div>
+          <ZoomOutButton wid={wid} loc={loc} node_loc={node_loc} />        </div>
```
johnchildren: Is this change intentional?

PR #280, opened 2026-01-14
Simplify Locs, remove -1s; symlink Input nodes in start_graph, and graph in each map-elem/loop-iter

acl-cqc:
* Add `start_graph` that takes `dict[string, OutputLoc]` and symlinks to those for all Input nodes in the graph (silently skipping any missing inputs)
* For Loc's ending in M or L, or the root Loc, do not `write_node_def`; instead `write_graph_def` (symlink).
   * `read_node_def` checks the Loc ends in N; `read_graph_def` the converse (returning a GraphData).
* Rename `NodeRunData` to `RunNodeTask` and define `Task = RunNodeTask | LoopIterTask` (returned in WalkResult) where the latter, used for `walk_loop`, calls `start_graph`
* Also add `body` input containing the graph, allowing recursive graphs
* Thus, remove all `N(-1)`s and "fake" `Eval((-1,..)...)`s, and hence the need for `ExteriorRef` in #266, which this is intended to precede

TODOs:
- [ ] Track Input nodes of Graph rather than search through all the nodes to find the Inputs?
- [ ] Use a better name than `body`? E.g. `__tkr_body`
- [x] Check visualizer still works

### DIFF
--- a/tierkreis_visualization/tierkreis_visualization/data/eval.py
+++ b/tierkreis_visualization/tierkreis_visualization/data/eval.py
@@ -4,7 +4,6 @@ from typing import assert_never
 from tierkreis.controller.data.core import NodeIndex
 from tierkreis.controller.data.location import Loc
 from tierkreis.controller.data.graph import GraphData, IfElse
-from tierkreis.controller.data.types import ptype_from_bytes
 from tierkreis.controller.storage.adjacency import in_edges
 from tierkreis.controller.storage.protocol import ControllerStorage
 
@@ -62,12 +61,12 @@ def add_conditional_edges(
         py_edges.append(edge)
 
 
-def get_eval_node(
-    storage: ControllerStorage, node_location: Loc, errored_nodes: list[Loc]
+def render_graph(
+    storage: ControllerStorage,
+    node_location: Loc,
+    graph: GraphData,
+    errored_nodes: list[Loc],
 ) -> PyGraph:
-    thunk = storage.read_output(node_location.N(-1), "body")
-    graph = ptype_from_bytes(thunk, GraphData)
-
     pynodes: list[PyNode] = []
     py_edges: list[PyEdge] = []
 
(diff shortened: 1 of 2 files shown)


tierkreis/tierkreis/controller/start.py
```diff
@@ -172,6 +170,32 @@ def start(
...
 
+def start_graph(
+    storage: ControllerStorage,
+    executor: ControllerExecutor,
+    loc: Loc,
+    graph_input: OutputLoc,
+    ins: dict[PortID, OutputLoc],
+) -> None:
+    # We have to write something here to mark the node/graph as started.
+    # TODO ALAN - pass NodeDef? But can't really make valid/correct inputs.
```
acl-cqc: This is a bit dubious, as per PR description

johnchildren (commented):
I didn't get into the meat of the PR yet, but I had a couple of comments about minor things

tierkreis/tests/controller/test_graphdata_storage.py
```diff
@@ -19,22 +18,42 @@
...
+    graph_def = storage.read_graph_def(loc)
+    assert graph_def.graph_inputs == inputs
+    assert set(graph_def.output_ports) == outputs
+
+
 @pytest.mark.parametrize(
     ["node_location_str", "graph", "port", "target"],
     [
-        ("-.N0", simple_eval(), "value", b"null"),
-        ("-.N4.M0", simple_map(), "0", b"null"),
```
acl-cqc: Not sure about this "0". Previous code did have something to explicitly add a "0" to the node outputs if there was a port `"*"`, which I have removed, because post-PR there wasn't a `"*"` either....

tierkreis/tests/controller/test_graphdata_storage.py
```diff
@@ -57,7 +76,7 @@ def test_raises() -> None:
     ["node_location_str", "graph", "target"],
     [
         ("-.N0", simple_eval(), ["value"]),
-        ("-.N4.M0", simple_map(), ["0"]),
```
acl-cqc: similarly

PR #290, opened 2026-01-20
chore: tierkreis riken starter

mwpb:
* (Make examples more consistent: use QASM file for example circuits.)
* Fix logging for pjsub executor.
  * Do not overwrite workflow logs file.
  * Use Bash to stream task output to workflow (and node) logs files. Omit `-o` and `-e` args to `pjsub`, which overwrites logs instead of appending.
  *  Stream the (small amount of) output from the subprocess call to both node and workflow logs files. This output usually just includes the pjsub Job ID.
  * The practical upshot of this is that the pjsub Job ID gets written as the first line of the node log file. (Pending future vis change this will be visible on click in the visualiser.) Also the Job ID gets printed to the workflow log file (but it is probably of less use there...).
* Create `TaskExecutor` that routes tasks to other executors based on the fully qualified task name. This means that different tasks in the same worker can be routed to different places.
* Importing workers from GitHub. This one is a bit speculative and could be separated out if we need more time to think. 
  * Define base class `TKRDependency` with a single `install` method. Implement subclass GitHubDependency.
  * Expose `install_dependencies`, `remove_dependencies` and `clear_cache` methods.
* Fix typo that prevented controller logs going into the correct place.
* Add some INFO level logging to aer_worker, mainly to see it appear separately in the different node log files.

### DIFF
--- a/tierkreis/tierkreis/pkg/base.py
+++ b/tierkreis/tierkreis/pkg/base.py
@@ -15,4 +15,4 @@ class TKRDependency(ABC, BaseModel):
 
     """Install a worker called `worker_name` into the `target_dir`.
     
-    Any cached files should be installed in the dir `worker_cache / self.type`."""
+    The `TKRDependency` has at its disposal a directory `worker_cache / TKRDependency.type` for any caching it needs."""


tierkreis/tierkreis/pkg/__init__.py
```diff
@@ -0,0 +1,21 @@
+from os import unlink
```
johnchildren: It's a bit unclear to me what this code does or why, can you add a docstring to the module or similar to explain what it is for and how it is intended to be used?

mwpb: Sure, will have a go this afternoon.

mwpb: Sure! Probably can do so this afternoon.

Issue #216, opened 2025-09-26
[Feat] Viz?: add a way to get data from checkpoints


philipp-seitz:
In some cases users might want to reuse intermediate results for further analysis:

Possible Solution:
- Make the value downloadable in the viz

Issue #196, opened 2025-09-12
Add Timestamps to nodes


philipp-seitz:
needs a draft for visuals

PR #297, opened 2026-01-23
Feat/vis improvements

philipp-seitz:
Updated the node info to address:
closes: #196 closes: #216 

Open for any visual suggestions

image

### DIFF
--- a/tierkreis_visualization/README.md
+++ b/tierkreis_visualization/README.md
@@ -1,15 +1,41 @@
 # Tierkreis visualizer
 
-WIP browser based visualizer.
-
-```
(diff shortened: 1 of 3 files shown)


mwpb (commented):
Thanks Philipp. Looks pretty reasonable to me. Just one comment about non-JSON outputs, which we should handle has gracefully as possible.

If you'd like to test such a situation then you can add an extra check

```python
try:
    return bs.decode()
except UnicodeDecodeError:
    return bs.decode("latin1")
```

to line 11 of `tierkreis_visualization/tierkreis_visualization/data/outputs.py` and look at the linspace node of the scipy graph in the examples folder. (This extra check should probably be present anyway...)

PR #309, opened 2026-01-29
fix: optional outputs in read_outputs 

mwpb:
* Modify `examples/defaults.py` so that its optional output is not provided.
* (Create helper function that determines if a Python type is optional.)
* (Encapsulate `FileNotFound` exception into an error generic over the storage.)
* Fix logs to explain that we will try to use default values where they are available.
* Fix `read_outputs` to understand that some values are optional. In case an output is optional then do not raise on `EntryNotFound`.

### DIFF
--- a/tierkreis/tests/controller/test_resume.py
+++ b/tierkreis/tests/controller/test_resume.py
@@ -4,7 +4,11 @@ from uuid import UUID
 
 import pytest
 
-from tests.controller.defaults_graphs import defaults_omit
+from tests.controller.defaults_graphs import (
+    defaults_not_none,
+    defaults_omit,
+    defaults_passthru,
+)
 from tests.controller.sample_graphdata import (
     maps_in_series,
     simple_eagerifelse,
@@ -30,6 +34,7 @@ from tests.controller.typed_graphdata import (
     gcd,
     typed_map_simple,
 )
+from tierkreis.builder import GraphBuilder
 from tierkreis.controller import run_graph
 from tierkreis.controller.data.types import PType
 from tierkreis.controller.executor.in_memory_executor import InMemoryExecutor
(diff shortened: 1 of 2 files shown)


tierkreis/tierkreis/controller/storage/exceptions.py
```diff
@@ -0,0 +1,9 @@
+from tierkreis.exceptions import TierkreisError
+
+
+class TierkreisStorageError(TierkreisError):
+    """An error with the chosen Tierkreis storage layer."""
+
+
+class EntryNotFound(TierkreisStorageError):
+    """Storage entry not found."""
```
acl-cqc: When you create this AFAICS you are passing in the `src` i.e. filename/path. I don't see that parameter/field here though?

examples/defaults.py
```diff
@@ -50,16 +50,20 @@ def inner_graph_2() -> GraphBuilder:
     return g
 
 
-def outer_graph() -> GraphBuilder:
+def outer_graph() -> GraphBuilder[OuterInputs, OuterOutputs]:
```
acl-cqc: just my understanding, it's not necessary to add these extra bits of annotation here? (But I agree one might as well)

acl-cqc: Ah, but *the* place to add such annotations is on `inner_graph_2' because that's the one that we are calling both with and without the default argument (EDIT:...which is working nicely, I might add 😁 👍 )

mwpb: Right!! I've put them on all builder functions in this file now...

mwpb:
Thanks Alan, I'm going to merge in as-is. If we decide against `examples/defaults.py` then we can easily remove that separately.

Issue #263, opened 2025-12-19
Improve Debug Data


philipp-seitz:
Currently debug data is a map: loop_name -> loc for the tracing.
We might want to improve this in the future

PR #325, opened 2026-02-12
Feat: debug data

philipp-seitz:
a couple of changes for better monitoring
closes: #263


tierkreis/tierkreis/controller/storage/protocol.py
```diff
@@ -94,6 +92,12 @@ def logs_path(self) -> Path:
     def debug_path(self) -> Path:
         return self.workflow_dir / "debug"
 
+    def _exec_data_path(self) -> Path:
```
mwpb: Hmm... I wonder if it would be easier for this method to take the `Loc` as an argument and avoid most of the logic in `append_executor_data`? If the lookup direction is node_loc -> data then I think we may as well separate them out into the directory representing the node_loc.

mwpb (commented):
Thanks Philipp! Some general suggestions and one specific one about `pjsub`. As a general comment, I'd be careful not to raise an Exception if something is not found (and that data is only used for the purposes of debugging). E.g. the Job ID. If there is a change on the HPC system, we'd rather not have to fix our debug data code urgently in addition to our runtime code.

mwpb (approved):
Thanks Philipp, those comments seem addressed well. One slight problem is that I appear to have lost access to fugaku. So it might be worth you doing a quick double check that the symbolic circuits example is working before merging....

PR #314, opened 2026-02-04
Docs: examples

philipp-seitz:
This rewrites the examples as jupyter notebooks
closes: #287

PR #335, opened 2026-02-18
chore: Use standard directory for example

johnchildren:
Revert change to symbolic circuits checkpoints directory


philipp-seitz:
I kept this as an example in #314 but after merging the examples wil be removed

Issue #367, opened 2026-02-27
[DOCS BUG] No explanation how to register (prepackaged) workers


rossduncan:
The docs on workers do not mention the need to register the workers with the executor.  I have looked in the following places 
* https://quantinuum.github.io/tierkreis/worker/index.html
* https://quantinuum.github.io/tierkreis/worker/native_workers/index.html
* https://quantinuum.github.io/tierkreis/worker/native_workers/pytket_worker.html

I guess these are the obvious places.  Since the init project set up includes an executor that works for custom workers this is a bigger issue for preprackaged workers.  I can't find anywhere in the docs a *working* example of using a prepackaged worker... e.g. 
* https://github.com/Quantinuum/tierkreis/blob/main/docs/source/examples/hamiltonian.ipynb
shows 
```
common_executor = UvExecutor(
    registry_path=PACKAGE_PATH.parent / "tierkreis_workers",
    logs_path=storage.logs_path,
)
```
but the registry path given does not exist in the default project.

Issue #360, opened 2026-02-26
Docs should describe the behaviour of this command


rossduncan:
It's non-obvious so it should be documented.

Issue #359, opened 2026-02-26
[BUG] tkr init worker creates spurious directories


rossduncan:
## Describe the bug
When we run `tkr init worker ...` it creates a directory structure `tkr/workers/workername` regardless of where it is in the current project.  This is only the correct behaviour if it is in the top-level of the project. 

## To Reproduce 
run `tkr init worker -n testworker` when not in the project root.

## Expected behaviour 
It should recognise where in the project tree it is and create the worker in the correct location.  A flag may be needed to reenable the dumb behaviour.

PR #379, opened 2026-03-03
feat: update workers and executors

philipp-seitz:
Adds the following:
# Workers (init)
- switch from (`tkr-worker-name`,`tkr-worker-name-api`) to (`tkr-worker-name`,`tkr-worker-name-impl`)
- make api a sibling of src
- separate main.py from task definitions 

# Executor
- allows the shell executor to run site packages

# Docs:
- update the docs on `tkr init worker` closing: #359 and #360 
- add docs no how to run site packages closing: #367 

# TODOs:
I haven't updated existing workers this will be in a separate pr

### DIFF
--- a/tierkreis/tierkreis/worker/worker.py
+++ b/tierkreis/tierkreis/worker/worker.py
@@ -209,10 +209,13 @@ class Worker:
         :param argv: The cli args.
         :type argv: list[str]
         """
-        if len(argv) <= 1:
-            print(self.name)
-            return
         handler = add_handler_from_environment(logger)
+        if len(argv) <= 1:
+            logger.warning("No arguments provided to invoke worker %s", self.name)
+            logger.warning(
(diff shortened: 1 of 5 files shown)


tierkreis/tierkreis/worker/worker.py
```diff
@@ -203,10 +203,15 @@ def app(self, argv: list[str]) -> None:
...
 
         Either generate stubs or run the worker.
+        If no arguments are passed immediately returns.
+        This is used to check the command exists.
 
         :param argv: The cli args.
         :type argv: list[str]
         """
+        if len(argv) <= 1:
+            print(self.name)
```
johnchildren: Do you still need this print?

philipp-seitz: Technically not, but it's usefull to verify the worker is doing what expected. If you think I should remove it I can do so

PR #381, opened 2026-03-04
GraphBuilder->Graph with separate Workflow type

acl-cqc:
* Add parametrized `Workflow` holding a GraphData and outputs_type.
* GraphBuilder renamed to Graph, and `.outputs` renamed to `.finish_with_outputs` which returns a Workflow
   - [ ] TODO: Rename `finish_with_outputs` to `build`??
* Change methods that take Graph(Builders) (e.g. `loop`, `eval`) to take Workflow's, removing check in `embed` that the Output node is present (this is now guaranteed if your python code is well-typed - I was surprised not to find more such checks in map/loop/eval/const/etc.)
* Add a `ModelConvertible` protocol like DictConvertible/etc. but that works for parametrized types. (This was a bit of a can of worms, there are some slightly suspect bits in `types.py`.) Workflow implements this, and serializes as a `GraphData`. (Type erasure: Workflow is a build-time-only construct).
   - [ ] This was in commit 6ccf0805ba1b4379f6d1d527ef6cf14cd3dbe55e, previously I included Workflow in the union of `PType`'s which was a bit hacky (special-case, awkward circular import, etc.) but a lot shorter/easier (and `DictConvertible` makes me worry about robustness and parametrization). Thoughts?
* Update stub generator to support TNamedModel's as well as PModel `Struct`s (by adding an `is_ptype` field to GenericType - this is not super-neat, suggestions of better ways to do this welcome, but my previous attempt was even (much) worse]: 454c18f370162e1296873f8b1accb68d23cd9e7d)
   * It'd be good to remove `@runtime_checkable` from the `RestrictedNamedTuple` subclasses/instantiations, but leaving that for [Another PR](https://github.com/Quantinuum/tierkreis/pull/401)
* Update test functions, docs/notebooks, etc., and test worker (including tests from #392) to take/return parametrized Workflow rather than GraphData. This was a bit tedious, and the bulk of changed files in the PR... I have tried to update the text of the docs a little but there is a lot of use of the word "graph" to mean either an unfinished one or a finished one, and trying to distinguish which of those should be workflows is...tricky. This may suggest calling the two classes `GraphBuilder` and `Graph` rather than `Graph` and `Workflow`???

The end result is that you need fewer `TypedGraphRef`s (but still some), and when you do need one, `pyright` should catch cases of providing the wrong `outputs_type` (because this is now just a runtime reification of a type that `pyright` knows statically).

Other possible TODOs (maybe too much for one PR??)
- [ ] Remove output type from Graph (and provide when you set the outputs)? This would mean splitting the class hierarchy i.e. we'd need a parent `_BaseGraph[Inputs]` with all methods *except* `finish_with_outputs` and `graph_ref`, then `class Graph[Inputs](_BaseGraph[Inputs]): ...fn finish_with_outputs[Outputs](self, outputs_type: type[Outputs], outputs: Outputs)` and separately `class RecursiveGraph[Inputs,Outputs](_BaseGraph[Inputs]): ...fn finish_with_outputs(self, outputs: Outputs)` plus `fn graph_ref(self)`. Not sure whether this is worth it so I think leave for another PR?
- [ ] We have quite a few tests that build GraphData's directly (not using the builder)....if we want the builder to be the only supported way of building graphs, we'll need to update these too.
   - [ ] If we do that, then this would allow hiding GraphData (i.e. making private, this is python) where it is used in the builder, so GraphData is really part of the graph runtime, not part of the user-facing graph-building API.

### DIFF
--- a/tierkreis/tests/workers/graph/main.py
+++ b/tierkreis/tests/workers/graph/main.py
@@ -29,7 +29,7 @@ def graph_of_graph(
     The graph contains the argument graph `f` as a constant."""
     g = Graph(TKR[int], TKR[int])
     v = g.inputs
-    ref: TypedGraphRef[TKR[int], TKR[int]] = g._graph_const(f)
+    ref: TypedGraphRef[TKR[int], TKR[int]] = g.graph_const(f)
     for _ in range(n):
         v = g.eval(ref, v)
     return g.finish_with_outputs(v)
--- a/tierkreis/tierkreis/builder.py
+++ b/tierkreis/tierkreis/builder.py
@@ -188,6 +188,8 @@ class Graph[Inputs: TModel, Outputs: TModel]:
     def const[T: PType](self, value: T) -> TKR[T]:
         """Add a constant node to the graph.
 
+        Note if the value is a Workflow, use `graph_const` instead to preserve type information.
+
         :return: The constant value.
         :rtype: TKR[T]
         """
@@ -251,11 +253,15 @@ class Graph[Inputs: TModel, Outputs: TModel]:
         )("value")
         return TKR(idx, port)
 
-    def _graph_const[A: TModel, B: TModel](
+    def graph_const[A: TModel, B: TModel](
         self,
         graph: Workflow[A, B],
     ) -> TypedGraphRef[A, B]:
-        # TODO @philipp-seitz: Turn this into a public method?
+        """Add a graph constant to the graph.
+        Returns a typed reference to the graph that can be used in eval and loop nodes.
+
+        :param graph: The graph to add as a constant.
+        """
         return TypedGraphRef(
             self.const(graph),
             graph.outputs_type,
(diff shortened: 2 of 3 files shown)


tierkreis/tests/workers/graph/stubs.py
```diff
@@ -1,13 +1,43 @@
 """Code generated from graph namespace. Please do not edit."""
 
-from typing import NamedTuple
-from tierkreis.controller.data.models import TKR, OpaqueType
+from typing import NamedTuple, Protocol
+from tierkreis.controller.data.models import TKR
+from tierkreis.controller.data.types import FinishedGraph
+
+
+# ALAN manually edited this, generated stub had `class ApplyTwiceInput(Struct, Protocol)`
```
acl-cqc: Leaving this as generated gives error "Can't instantiate protocol" from typed_graphdata.py; giving that it's own version (like this) and leaving the generated version here passes tests but fails pyright

PR #391, opened 2026-03-04
refactor: remove unused generics_in_ptype/model, put TypedGraphRef.inputs_type before outputs_type

acl-cqc:
Also remove some overloads that seem pointless (??), and tidy some imports

Removing generics_in_ptype as a preliminary to #381 because it's hard to handle parametrized FinishedGraph.

`TypedGraphRef[Inputs, Outputs]` was constructed via `TypedGraphRef(outputs_type, inputs_type)`, so reorder the latter. (Although the `inputs_type` field is unused it helps `pyright` figure out the `Inputs` type parameter when in strict mode or with `reportUnknownVariableType` so keep it for now.)

### DIFF
--- a/tierkreis/tierkreis/builder.py
+++ b/tierkreis/tierkreis/builder.py
@@ -70,8 +70,8 @@ class TypedGraphRef[Ins: TModel, Outs: TModel]:
     """
 
     graph_ref: ValueRef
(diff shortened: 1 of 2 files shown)


mwpb (approved):
This looks like a good tidy up, thank you.

PR #395, opened 2026-03-05
chore: improve qol (docs+cli)

philipp-seitz:
depends on some changes in #386

### DIFF
--- a/tierkreis/tests/cli/test_tkr.py
+++ b/tierkreis/tests/cli/test_tkr.py
@@ -98,7 +98,7 @@ default_args = [
 
 cli_params = [
     (
-        [*default_args, "-f", "tierkreis/tests/cli/data/sample_graph"],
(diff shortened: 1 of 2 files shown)


tierkreis/tierkreis/cli/project.py
```diff
@@ -219,6 +220,8 @@ def run_args(args: argparse.Namespace) -> None:
         graphs_dir.mkdir(exist_ok=True, parents=True)
         with Path.open(graphs_dir / "main.py", "w+", encoding="utf-8") as fh:
             fh.write(default_graph(worker_name))
+        with Path.open(graphs_dir / "values.json", "w+", encoding="utf-8") as fh:
```
johnchildren: I'm not certain if this is the best location for the template values, I would perhaps suggest the project root directory instead but I guess we might also want to encourage people to run workflows with multiple different sets of inputs?

Issue #204, opened 2025-09-19
Look into PSI/J as Executor or extend HPC executors


philipp-seitz:
We should look at this as an alternative to the HPC executors

Issue #182, opened 2025-09-08
Chore: Reenable HPC executors once thoroughly tested


philipp-seitz:
PBS and PSJUB executors are currently disabled due to the lack of testing

PR #426, opened 2026-04-09
feat: psij executors

philipp-seitz:
creates executors with psij, still needs some more testing


philipp-seitz:
added a pbs container so we now can fully test it.
PSIJ is working too.
Closes: #182

infra/pbs_local/Dockerfile
```diff
@@ -0,0 +1,135 @@
...
+       expat-devel \
+       openssl-devel \
+       libXext \
+       libXft \
+       autoconf \
+       automake \
+       gcc-c++ \
+       git \
+       wget \
+    && yum -y install \
```
johnchildren: I don't know `yum` that well, but I'm curious why you need two separate install commands. Possibly the second set depend on the first set? Did you get this list of packages from the pbs docs?

tierkreis/tests/executor/test_hpc_executor.py
```diff
@@ -25,35 +26,68 @@ def mpi_graph() -> GraphData:
...
 
-def job_spec() -> JobSpec:
+def slurm_spec() -> JobSpec:
     return JobSpec(
         job_name="test_job",
         account="test_usr",
-        command=(
-            "--allow-run-as-root /root/.local/bin/uv run /slurm_mpi_worker/main.py "
-        ),
+        command=("--allow-run-as-root /root/.local/bin/uv run /mpi_worker/main.py "),
```
johnchildren: I realise it predates this PR, but I've seen this `--allow-run-as-root` in a few places, what is the purpose of the flag and what does it get us?

tierkreis/tierkreis/controller/executor/hpc/hpc_executor.py
```diff
@@ -89,8 +89,7 @@ def run_hpc_executor(
         executor.logs_path,
         executor.errors_path,
         spec.command,
-    )
-
+    )[1:-1]  # remove the enclosing quotes
```
johnchildren: This seems slightly perverse way to strip the command when .strip exists and wouldn't require a comment:

```python
>>> "\"hello\""
'"hello"'
>>> "\"hello\"".strip("\"")
'hello'
```

But why does add_std_handlers return a string that contains quotes anyway?

tierkreis/pyproject.toml
```diff
@@ -14,6 +14,7 @@ authors = [
 
 requires-python = ">=3.12"
 dependencies = [
+    "psij-python>=0.9.11",
```
johnchildren: Not important right now, but I wonder if we should make this an optional extra?

infra/pbs_local/main.py
```diff
@@ -0,0 +1,53 @@
...
+from sys import argv
+
+from mpi4py import MPI  # type: ignore
+
+from tierkreis import Worker
+
+logger = logging.getLogger(__name__)
+worker = Worker("mpi_worker")
+
+comm = MPI.COMM_WORLD
```
johnchildren: What is the purpose of this global `comm` variable if you invoke it later in the various functions anyway?

infra/pbs_local/main.py
```diff
@@ -0,0 +1,53 @@
...
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
+    hostname = socket.gethostname()
+    return {"rank": rank, "hostname": hostname}
+
+
+@worker.task()
+def mpi_rank_info() -> str | None:
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
```
johnchildren, line 22: You are currently getting the rank twice with _proc_info as well

infra/slurm_local/sbatch
```diff
@@ -1,15 +1,9 @@
...
+SCRIPT_FILE=$(basename "$1")
 
-NODE_PATH=$(sed -n -E '$ s#.*([0-9a-fA-F-]{36}/[^/]+)/definition$#\1#p' "$5")
-WORKFLOW_ID=$(echo "$NODE_PATH" | cut -d'/' -f1)
-docker cp "$5" slurmctld:/data/"$SCRIPT_FILE"
-docker exec slurmctld sh -c 'rm -rf /data/*.log && \
-ln -s $HOME/.tierkreis/checkpoints/$2/logs /data/logs.log && \
-ln -s $HOME/.tierkreis/checkpoints/$3/errors /data/errors.log  && \
-sbatch --output=logs.log --error=errors.log --chdir=/data /data/$1' sh "$SCRIPT_FILE" "$WORKFLOW_ID" "$NODE_PATH"
+# This emulates the sbatch command assuming .tkr is in $HOME
```
johnchildren: what is `.tkr`? the `.tierkreis` directory?

tierkreis/tierkreis/controller/executor/hpc/psij_executor.py
```diff
@@ -0,0 +1,95 @@
...
+        self.errors_path = logs_path
+        self.spec = spec
+        self.psij_executor = psij_executor
+
+    def run(
+        self,
+        launcher_name: str,
+        worker_call_args_path: Path,
+    ) -> ExecutorDebugData:
+        self.errors_path = (
```
johnchildren, line 49: overwriting the errors path on the object seems a bit weird here? It this pattern used in the other executors? Indeed it even ends with `logs` ??

tierkreis/tierkreis/controller/executor/hpc/psij_executor.py
```diff
@@ -0,0 +1,95 @@
...
+        job = Job(self.spec)
+        self.spec.executable = exec_file.name
+
+        try:
+            self.psij_executor.submit(job)
+            with Path.open(self.logs_path, "a+") as fh:
+                fh.write(str(job.status))
+
+            with Path.open(self.errors_path, "a+") as fh:
+                fh.write(str(job.status))
```
johnchildren, line 92: Why do we write this to both paths ooi?

tierkreis/tierkreis/controller/executor/hpc/psij_executor.py
```diff
@@ -0,0 +1,95 @@
...
+        tkr_dir = Path(self.spec.environment[TKR_DIR_KEY])
+        with tempfile.NamedTemporaryFile(
+            mode="w", dir=str(tkr_dir), prefix="tkr_eval_", suffix=".sh", delete=False
+        ) as exec_file:
+            exec_file.write('#!/bin/bash\neval "$@"\n')
+            exec_file.close()
+            os.chmod(exec_file.name, 0o755)
+
+        job = Job(self.spec)
+        self.spec.executable = exec_file.name
```
johnchildren: As a point of style I'm not a huge fan of mutating variables on the class that only apply to the current invocation of the function. It's also slightly surprising behavior that we mutate the Job itself via mutating attribute on the class.

I think it would be clearer to deepcopy the spec into the Job object or similar.

tierkreis/tests/executor/test_hpc_executor.py
philipp-seitz: The slurm container is not set up with users beside root. By default MPI disables running as root you can either set the flag or two env vars. This is only relevant for testing

tierkreis/tierkreis/controller/executor/hpc/psij_executor.py
philipp-seitz: This is indeed a pattern that we repeat in other executors currently.
`self.errors_path` should probably be now `self.node_logs`.
This is from the `WorkerCallArgs` where it was called errors originally.
Happy to change this in a different pr

tierkreis/tierkreis/controller/executor/hpc/psij_executor.py
philipp-seitz: we currently use `self.logs_path=<workflow_id>/logs` as central log for everything while `self.errors_path=<workflow_id>/<node_id>/logs` is node specific. The latter we use for error informartion e.g. in the visualizer and error reporting

philipp-seitz:
also closes: #204

tierkreis/tierkreis/controller/executor/hpc/psij_conversion.py
```diff
@@ -0,0 +1,195 @@
...
+        process_count=spec.mpi.proc if spec.mpi else None,
+        processes_per_node=spec.mpi.max_proc_per_node if spec.mpi else None,
+        cpu_cores_per_process=spec.resource.cores_per_node,
+        gpu_cores_per_process=spec.resource.gpus_per_node,
+        memory=spec.resource.memory_gb * 1024 * 1024 * 1024
+        if spec.resource.memory_gb
+        else None,
+    )
+
+    # 2. Map Attributes and capture custom fields that psij doesn't natively support
```
johnchildren: is there a 1. or 3. comment somewhere?

docs/source/executors/hpc/psij_executors.md
```diff
@@ -0,0 +1,91 @@
...
+For most arguments `psij.JobSpec` and `tierkreis.JobSpec` and their respective executors overlap.
+Here we list the known cases where they are not congruent:
+
+### Executors
+
+The main difference is:
+- Tierkreis uses scheduler specific executor classes
+- You cannot provide a path to the scheduler command binary for `psij`. The command has to be in `$PATH`.
+- The `launchers_path` argument has no effect on the `PSIJExecutor`. We reommend using fully qualified paths for commands, e.g. `/home/pbsuser/.local/bin/uv run /mpi_worker/main.py` (for running on the compute nodes).
+- `psij` uses an internal script to launch jobs, this can lead to incompatibilities when using custom
```
johnchildren: custom what?

johnchildren (approved):

PR #433, opened 2026-04-17
Rust orchestrator

johnchildren:


tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn outputs(self) -> Option<HashMap<String, AssetSpec>> {
+        match self.status {
+            Status::Complete { outputs } => Some(outputs),
+            _ => None,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Status {
```
philipp-seitz, line 30: Probably overkill, but what about other states, e.g. created?
Is there a way of indicating the number of retries here / somewhere else

tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub outputs: HashSet<String>,
+    // Override the default from the executor.
+    pub output_storage_name: Option<String>,
+
+    pub resources: HashMap<String, Value>,
+    pub environment: HashMap<String, Value>,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct WorkerSpec {
```
philipp-seitz, line 53: might be useful to know if a worker is internal (aka from python with `@worker.task`) or external (a binary we don't control).
Also have a mention about the expected behavior of workers: binary worker_cal_args etc.

tierkreis/src/asset_storage/interface.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub fn path(&self) -> miette::Result<PathBuf> {
+        match &self.kind {
+            AssetKind::File { root: parent } => Ok(parent.join(self.asset_key.0.to_string())),
+            _ => Err(miette!("Cannot build a path from a non-File asset!")),
+        }
+    }
+}
+
+/// [`AssetKey`] is a unique key for storing Assets.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
philipp-seitz: what does the order imply for asset keys? is it just for lexicographical sorting?

johnchildren: I was originally using them in a BTreeMap where ordering is important, but that is no-longer the case and I can just remove that now

tierkreis/src/asset_storage/interface.rs
```diff
@@ -0,0 +1,134 @@
...
+impl Display for AssetKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// The [`AssetStorage`] defines the minimum methods required for Assets to be stored.
+///
+/// The interface is essentially a key-value store, keyed by [`AssetKey`]s.
+pub trait AssetStorage: Send + Sync {
```
philipp-seitz, line 112: something like `invalidate` might be useful in the future

tierkreis/src/asset_storage/interface.rs
johnchildren: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html

tierkreis/src/asset_storage/interface.rs
johnchildren: Makes sense, I guess we may also want to be able to do something like specify retention periods or similar?

tierkreis/src/asset_storage/inmemory.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::asset_storage::interface::{AssetKey, AssetKind, AssetStorage};
+
+/// [`InMemoryStorage`] is an implementation of [`AssetStorage`] that stores
+/// Assets in a concurrent map data structure using [`AssetKey`]s as keys.
+#[derive(Clone, Debug)]
+pub struct InMemoryStorage {
+    // DashMap is a concurrent HashMap that lets us avoid locking the entire
+    // storage when saving/loading values.
+    store: DashMap<AssetKey, Value>,
```
philipp-seitz: this restricts us to json values. Need to consider custom serialized types, e.g. hugrs

johnchildren: Indeed, this was just a bit of a placeholder until I could replace it with a proper Value type, sorry I should have included a comment about that.

tierkreis/src/event.rs
```diff
@@ -0,0 +1,65 @@
...
+/// [Event] messages are emitted from [Executor][crate::executor::Executor] instances and the
+/// [Orchestrator][crate::orchestrator::Orchestrator].
+///
+/// [Event] messages correspond to an update in the state of a Node during the Workflow
+/// execution.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Event {
+    /// The id of an executor subprocess
+    // TODO: This should really be the id of a node instead, probably
+    // something like Loc which is programmatically determined at
```
philipp-seitz, line 21: Locs are not unique though, so rather somthing like workflowid+runid+loc

johnchildren: Sure, I guess it depends on the lifecycle we expect for the Executors as well, but this is slightly related to what I was talking about in slack and how many details we want to pass down.

tierkreis/src/server.rs
```diff
@@ -0,0 +1,66 @@
+/*!
+The server module defines the HTTP interface to the Workflow server.
+*/
```
philipp-seitz, line 3: Is this intended to replace the fastapi one?
Some of the changes (e.g asset storages, runs) might also be relevant for changes in the frontend.

johnchildren: I think longer term yes, I realise there will probably be a few difference we will need to handle

tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct TaskPlan {
+    /// The name of the Worker to invoke.
+    pub worker_name: String,
+    /// The name of the Worker's task to invoke.
+    pub task_name: String,
+    /// The names and locations of the inputs to the task.
+    pub inputs: HashMap<String, AssetSpec>,
+    /// The names of the outputs of the task.
+    // TODO: We can find this out from the workers
```
philipp-seitz, line 24: but might be the case that we don't need all of the outputs the worker declares

tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub outputs: HashSet<String>,
+    /// A specific [`AssetStorage`][crate::asset_storage::AssetStorage]
+    /// from an [`AssetStorageRegistry`][crate::asset_storage::AssetStorageRegistry]
+    /// to save outputs to.
+    ///
+    /// If not specified the [Executor] will save the outputs to a default
+    /// [`AssetStorage`][crate::asset_storage::AssetStorage] instead.
+    pub output_storage_name: Option<String>,
```
philipp-seitz, line 32: implies that all task outputs will be written to the same storage.
Thinking of the ntchem example file outputs could be different from the return value of the binary

tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub environment: HashMap<String, Value>,
+}
+
+/// [`WorkerSpec`] defines the information about a Worker returned by an
+/// [Executor] from the [`Executor::workers`] method.
+#[derive(Debug, PartialEq)]
+pub struct WorkerSpec {
+    /// The name of the Worker.
+    pub worker_name: String,
+}
```
philipp-seitz, line 47: I assume this is tbd.? you mention meta data further down but I guess it's supposed to be more than the name? At least in the current design, could point to the api/give an appreviated version of it

tierkreis/src/executor/inmemory.rs
```diff
@@ -0,0 +1,634 @@
...
+};
+
+/// [`InMemoryResourceSpec`] determines what Resources should be available to the
+/// [`InMemoryExecutor`] or what is requested as part of a [`TaskPlan`].
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct InMemoryResourceSpec {}
+
+/// [`InMemoryEnvironmentSpec`] determines the default execution environment of
+/// [`InMemoryExecutor`] or what is requested as part of a [`TaskPlan`].
+// TODO: It's perhaps a bit unclear what this means for in-memory execution?
```
philipp-seitz, line 38: I'm not sure I understand. I guess this would be something like available modules?
But if a worker is installed this would mean everything should be acaialble, else how would it run?
And I assume we don't target stuff like numa information

tierkreis/src/executor/inmemory.rs
```diff
@@ -0,0 +1,634 @@
...
+                        .try_send(Event {
+                            id,
+                            status: Status::Cancelled {},
+                        })
+                        .expect("Failed to send update");
+                }
+            }
+            Some(res) = running.next() => {
+                // Ignore JoinError
+                let Ok((id, output_storage_name, outputs)) = res else { continue };
```
philipp-seitz: I dont fully understand this

tierkreis/src/executor/interface.rs
johnchildren: Indeed, I would like to flesh this out with information about which tasks are available and what their signatures are.

johnchildren: It kind of links in with trying to improve the IDL as well

tierkreis/src/executor/inmemory.rs
johnchildren: It's just ignoring the error, I think I can probably make it simpler

johnchildren: (and possibly just not ignore the error)

tierkreis/src/executor/interface.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// The [Executor] defines the minimum methods required for Task execution.
+pub trait Executor: Send + Sync {
```
philipp-seitz, line 50: I think we will need something like `register_worker`

tierkreis/src/executor/subprocess.rs
```diff
@@ -0,0 +1,713 @@
...
+pub struct SubprocessEnvironmentSpec {}
+
+/// [`SubprocessExecutor`] defines an [Executor] that performs Task Nodes using Worker subprocesses.
+pub struct SubprocessExecutor {
+    event_sender: mpsc::Sender<Event>,
+    event_receiver: Mutex<Option<mpsc::Receiver<Event>>>,
+    cancel_senders: Mutex<HashMap<u32, oneshot::Sender<()>>>,
+
+    // The name of the storage that the subprocess will read
+    // and write files from. Must be file based.
```
philipp-seitz, line 184: must not be in memory based (unless we allow MPI ranks, not sure if its a useful exercise)

tierkreis/src/executor/subprocess.rs
```diff
@@ -0,0 +1,713 @@
...
+        asset_storage_registry: &AssetStorageRegistry,
+        subprocess_storage_name: &str,
+        output_storage_name: &str,
+    ) -> miette::Result<Self> {
+        let asset_storage_registry_lock = asset_storage_registry
+            .read()
+            .map_err(|err| miette!("Failed to lock AssetStorageRegistry for reading: {err}"))?;
+        if !asset_storage_registry_lock.contains_key(subprocess_storage_name) {
+            return Err(miette!("subprocess_storage_name not in registry"));
+        }
```
philipp-seitz: we could check the kind right?

tierkreis/src/executor/subprocess.rs
johnchildren: I guess this more of an implementation detail that our worker protocol requires input files, I think changing this would require changes to the worker protocol?

tierkreis/src/executor/subprocess.rs
johnchildren: right we should, that's a mistake by me

tierkreis/src/executor/subprocess.rs
```diff
@@ -0,0 +1,713 @@
...
+                        }
+                    }
+                }
+            };
+
+            tasks.push(task);
+        }
+
+        // This guarantees the tasks will complete in order as far as the runtime
+        // is concerned.
```
philipp-seitz: is this a necessary requrirement? I thought only the dispatch order matters (i.e. all inputs must be ready)

johnchildren: It is not a requirement and I'm also pretty inconsistent about it, I guess ideally we might want to pass the task dependencies into the executor as well but I didn't think enough about the correct way to do this.

tierkreis/src/executor/interface.rs
johnchildren: Could be useful, but it might depend a bit about how long we expect Executor objects to live, something to discuss in person tomorrow perhaps?

tierkreis/src/executor/subprocess.rs
```diff
@@ -0,0 +1,713 @@
...
+/// [`SubprocessEnvironmentSpec`] determines the default execution environment of
+/// [`SubprocessExecutor`] or what is requested as part of a [`TaskPlan`].
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct SubprocessEnvironmentSpec {}
+
+/// [`SubprocessExecutor`] defines an [Executor] that performs Task Nodes using Worker subprocesses.
+pub struct SubprocessExecutor {
+    event_sender: mpsc::Sender<Event>,
+    event_receiver: Mutex<Option<mpsc::Receiver<Event>>>,
+    cancel_senders: Mutex<HashMap<u32, oneshot::Sender<()>>>,
```
philipp-seitz: whats the reason this is oneshot and not mpsc?

johnchildren: There's a slight inconsistency in the way I implemented the subprocess and the inmemory executors that I can probably fix where the subprocess executor futures are just spawned into the background and they handle their own cancellation, whereas the inmemory one has a central cancellation channel. As such cancellation is performed with a oneshot channel for the subprocess one as only one message needs to be sent.

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+        let asset_storage_registry_lock = asset_storage_registry
+            .read()
+            .map_err(|err| miette!("Failed to lock AssetStorageRegistry for reading: {err}"))?;
+        if !asset_storage_registry_lock.contains_key(default_storage_name) {
+            return Err(miette!("default_storage_name not in registry"));
+        }
+
+        if !executor_registry.contains_key(default_executor_name) {
+            return Err(miette!("default_executor_name not in registry"));
+        }
```
philipp-seitz, line 71: depending how often we need this we could change this to a function on the registry?

johnchildren: Sure, right now it just a type alias, but we could wrap it up into a struct if there's a small number of methods that we use like this.

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+
+                    inputs,
+
+                    output_storage_name: Some(self.default_storage_name.clone()),
+                    ..Default::default()
+                }];
+                let default_executor_name = &self.default_executor_name;
+                let executor = self
+                    .executor_registry
+                    .get(&self.default_executor_name)
```
philipp-seitz, line 113: do I understand correctly that we always choose the same executor?

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+                event_sender
+                    .try_send(Event {
+                        id: 0,
+                        status: Status::Complete { outputs },
+                    })
+                    .into_diagnostic()?;
+            }
+            Node::Output {} => {
+                // Notify that the outputs are ready
+                let mut event_sender = self.event_sender.clone();
```
philipp-seitz, line 145: who is consuming this?

tierkreis/src/orchestrator.rs
johnchildren: Currently yes, I would like to be able to set the executor on the node but I am not sure what the best way to do this is. As in, should be metadata or similar?

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+                event_sender
+                    .try_send(Event {
+                        id: 0,
+                        status: Status::Complete { outputs },
+                    })
+                    .into_diagnostic()?;
+            }
+            Node::Eval {} => {
+                // spawn a sub-orchestrator?
```
philipp-seitz, line 182: I don't think this one possible way but then we need to deal with nested orchestrators

tierkreis/src/orchestrator.rs
johnchildren: All of the events end up in the stream provided by `listen`. Ideally I want to then use this stream to build up the execution state somewhere which then drives the next round of execution.

tierkreis/src/orchestrator.rs
johnchildren: Yeah I've been thinking about this today and I think we want to avoided nested orchestrators but it needs some thinking about how we handle locations.

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+            &registry,
+            &orchestrator.default_storage_name,
+            &events[1].clone().outputs().unwrap(),
+            json!({"value": 4}),
+        );
+
+        Ok(())
+    }
+
+    // Test that we can launch a single task and listen for
```
philipp-seitz: node/input

johnchildren: ah yes I forgot to update the docstrings on almost every test in that module, apologies

tierkreis/src/orchestrator.rs
```diff
@@ -0,0 +1,519 @@
...
+    // the status changes.
+    #[rstest]
+    #[case("memory")]
+    #[case("file")]
+    #[tokio::test]
+    async fn start_simple_workflow(#[case] default_storage_name: &str) -> miette::Result<()> {
```
philipp-seitz, line 412: some of this logic here should be wrapped into what it is now the controller, right?

johnchildren: right, ideally I would like to try and implement the same functionality as the controller so we can start re-using the existing test suite

tierkreis/src/executor/interface.rs
johnchildren: It's something we could do later, I just did the easiest thing

johnchildren: It also doesn't need to be optional - we do always know what the output storage name should be at this point

tierkreis/src/asset_storage/inmemory.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/433/changes/42282ea5cc44c0e6307b9ec2d6355a4a0058113a resolved here

tierkreis/src/executor/subprocess.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/433/changes/2b84d90685544c242a2c1f55dcda0468ea6ff30b hopefully fixed here

tierkreis/src/executor/inmemory.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/433/changes/81f9a204ac9f2c0aba413770ddaeaaf28312c8d9 hopefully improved here. As we don't have the task id we can't send an error so we have to panic for now.

tierkreis/src/asset_storage/interface.rs
johnchildren: Removed the extra derives

tierkreis/src/event.rs
johnchildren: Discussed in person

tierkreis/src/server.rs
johnchildren: Discussed in person

tierkreis/src/executor/subprocess.rs
johnchildren: discussed in person

tierkreis/src/orchestrator.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/433/changes/71c2c709fcefb53d2ced9af9b51bb2ccc2981b9c fixed here

tierkreis/src/orchestrator.rs
johnchildren: discussed in person

tierkreis/src/asset_storage/file.rs
```diff
@@ -50,21 +50,22 @@ impl AssetStorage for FileAssetStorage {
...
-        serde_json::to_writer(&mut file, &value).into_diagnostic()?;
+        file.write_all(&value).into_diagnostic()?;
         Ok(())
     }
 
-    fn load(&self, key: &AssetKey) -> miette::Result<Value> {
+    fn load(&self, key: &AssetKey) -> miette::Result<Vec<u8>> {
         let location = self.location(key);
-        let file = File::open(self.location(key))
+        let mut file = File::open(self.location(key))
```
philipp-seitz, line 64: why is this mut now?

johnchildren: Because the `read` trait needs it to change where it is looking at in the file, I think before we were passing ownership to serde_json which was doing it's own mutation

philipp-seitz (approved):

PR #457, opened 2026-05-05
feat: Refactor and expand orchestrator

johnchildren:
* Split Node into Action and NodeDefinition that describe an action
the Orchestrator has to perform versus the definition of a Node in
the graph.
* Add a new WorkflowGraph type that can be evaluated to find Nodes
to run.
* Add a new Execution Context that can be used with the WorkflowGraph
to plan a series of Actions.

### DIFF
--- a/tierkreis/src/graph.rs
+++ b/tierkreis/src/graph.rs
@@ -2,31 +2,220 @@
 This module defines the Workflow graph representation.
 */
 
-use serde_json::Value;
+use std::collections::HashMap;
 
-/// [Node] is a placeholder enum for the kinds of operations that can appear
-/// in a Workflow graph. To be replaced with something compatible with the
-/// existing python code.
-pub enum Node {
-    /// An operation performed by a Worker.
-    Task {
-        /// The name of the Worker to call.
-        worker_name: String,
-        /// The name of the Task to call.
-        task_name: String,
-    },
-    /// An input to the Workflow.
+use bitvec::vec::BitVec;
+use miette::{Context, IntoDiagnostic, miette};
+use portgraph::{
+    Direction, LinkMut, LinkView, NodeIndex, PortGraph, PortIndex, PortMut, PortView,
+    algorithms::{TopoSort, toposort_filtered},
+};
+use serde::{Deserialize, Serialize};
+
+/// Possible definitions for Nodes in the [`WorkflowGraph`].
+#[derive(Debug, Serialize, Deserialize)]
+pub enum NodeDefinition {
+    /// A node that takes input values to the Graph.
     Input {
-        /// The name of the Workflow input to capture.
+        /// The name of the input value to the Graph.
         name: String,
     },
-    /// The output of the Workflow
+    /// A node that returns output values from the Graph.
     Output {},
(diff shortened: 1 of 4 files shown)


tierkreis/src/graph.rs
```diff
@@ -2,31 +2,220 @@
...
+
+/// The [`WorkflowGraph`] defines a Workflow that can be evaluated
+/// by the Tierkreis runtime.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WorkflowGraph {
+    graph: PortGraph,
+    node_definitions: HashMap<NodeIndex, NodeDefinition>,
+    port_indices: HashMap<NodeIndex, HashMap<String, PortIndex>>,
+    port_names: HashMap<PortIndex, String>,
+    output_node: NodeIndex,
```
philipp-seitz, line 49: migth be useful to have acccess to the input nodes too as in graphdata

tierkreis/src/orchestrator.rs
```diff
@@ -23,9 +25,43 @@ use crate::{
...
-    graph::Node,
+    graph::{NodeDefinition, WorkflowGraph},
 };
 
+static EMPTY_INPUTS: LazyLock<HashMap<String, AssetSpec>> = LazyLock::new(HashMap::new);
+
+/// [Action] is a placeholder enum for operation the Executor can perform
+/// when interpreting a Workflow graph.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Action {
```
philipp-seitz, line 36: I'm not sure I fully understand the use case of this.
Isn't in most cases the action a task for an executor? Else this is just a copy of the node type?
I would expect things like `DISPATCH` (eager) `COLLECT`, or `SCHEDULE`

tierkreis/src/orchestrator.rs
```diff
@@ -82,111 +118,181 @@ impl Orchestrator {
...
+                            Action::Task {
+                                worker_name: worker_name.clone(),
+                                task_name: task_name.clone(),
+                            },
+                            inputs,
+                        )]
+                    }
+                    NodeDefinition::Eval {} => {
+                        let inputs = collect_inputs(context, workflow_graph, n).unwrap();
+                        let graph_asset_spec = inputs.get("thunk").unwrap();
```
philipp-seitz: I think in the current graph data this is named body/graph and not thunk

tierkreis/src/orchestrator.rs
```diff
@@ -82,111 +118,181 @@ impl Orchestrator {
...
-
-                // Move the value into storage if needed.
-                if value_spec.storage_name != self.default_storage_name {
+                }),
+                Action::Input { name } => {
+                    // Assume the inputs are the inputs to the graph.
+                    //
+                    // Just pull out the named input and assign it to the "value" output.
+                    let value_spec = inputs
+                        .get(&name)
```
philipp-seitz, line 220: We also allow partial inputs to graphs currently so we could have default values for some things

philipp-seitz (approved):

philipp-seitz (approved):

PR #466, opened 2026-05-12
feat: Add rudimentary Location struct

johnchildren:
Changes the id field in events to a Location struct that allows us to keep track of where in the workflow graph an event occured.

### DIFF
--- a/tierkreis/src/event.rs
+++ b/tierkreis/src/event.rs
@@ -134,7 +134,6 @@ pub async fn send_complete(
     outputs: HashMap<String, AssetSpec, RandomState>,
 ) -> miette::Result<()> {
     let workflow_complete = loc.is_root();
-    dbg!(&loc);
     event_sender
         .send(Event {
             loc: loc.clone(),
--- a/tierkreis/src/event.rs
+++ b/tierkreis/src/event.rs
@@ -8,7 +8,7 @@ use std::{collections::HashMap, hash::RandomState};
 use futures::{SinkExt, channel::mpsc};
 use miette::miette;
 
-use crate::asset_storage::interface::AssetSpec;
+use crate::{asset_storage::interface::AssetSpec, location::Location};
 
 /// [Event] messages are emitted from [Executor][crate::executor::Executor] instances and the
 /// [Orchestrator][crate::orchestrator::Orchestrator].
@@ -17,11 +17,8 @@ use crate::asset_storage::interface::AssetSpec;
 /// execution.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Event {
-    /// The id of an executor subprocess
-    // TODO: This should really be the id of a node instead, probably
-    // something like Loc which is programmatically determined at
-    // runtime.
-    pub id: u32,
+    /// The location of the Node for this Event.
+    pub loc: Location,
     /// The new status of the Node.
     pub status: Status,
 }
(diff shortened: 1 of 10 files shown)


tierkreis/src/graph.rs
```diff
@@ -97,6 +97,11 @@ impl WorkflowGraph {
         self.node_definitions.get(&node)
     }
 
+    /// Iterate over the indices of all the nodes in the grraph.
```
philipp-seitz: grr 😠

tierkreis/src/event.rs
```diff
@@ -116,16 +130,24 @@ pub async fn send_cancelled(event_sender: &mut EventSender, id: u32) -> miette::
 /// Will return Err if the channel for `event_sender` is full or closed.
 pub async fn send_complete(
     event_sender: &mut EventSender,
-    id: u32,
+    loc: Location,
     outputs: HashMap<String, AssetSpec, RandomState>,
 ) -> miette::Result<()> {
+    let workflow_complete = loc.is_root();
+    dbg!(&loc);
```
philipp-seitz: dbg

tierkreis/src/location.rs
```diff
@@ -0,0 +1,68 @@
...
+/// order tierkreis graph.
+///
+/// A [`Location`] consists of a path of components that point to either:
+///
+/// * The "root" of the Graph itself (if the path is empty).
+/// * A specific node in a Graph or a Subgraph inside a higher order node.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default)]
+pub struct Location(Vec<LocationComponent>);
+
+impl Location {
```
philipp-seitz, line 24: I'm not sure I fully understand the intention for how to use this yet.
Is the root loc (vec!([])) the equivalent to the top level eval we had before? 
Eval would then be the only node that appends to a loc?
In that scenario we would lose the information of node type in the loc which I guess is only important to distinguish higher order nodes

philipp-seitz (approved):

tierkreis/src/location.rs
```diff
@@ -0,0 +1,68 @@
...
+This module defines the [`Location`] struct that is used throughout the Tierkreis
+runtime to specify the place in a Workflow graph that something has happened.
+*/
+use portgraph::NodeIndex;
+
+/// A component of the path for a [`Location`].
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum LocationComponent {
+    /// The [`NodeIndex`] of a Node in a Graph or Subgraph inside a higher order node.
+    Node(NodeIndex),
```
philipp-seitz, line 11: would it make sense to have multiple variants (e.g Eval(NodeIndex), Map(NodeIndex), ...)?

johnchildren: Yeah my plan is to add them later as needed, though importantly Map and Loop don't use NodeIndex, they use an element index and iteration variable.

tierkreis/src/graph.rs
johnchildren: I will blame the jetlag

tierkreis/src/location.rs
johnchildren: Well I guess Loop would be more like `Loop(NodeIndex, u64)`

tierkreis/src/event.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/466/changes/39c31b685c690f4b2bf8b0e09684a751363529a7

tierkreis/src/graph.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/466/changes/39c31b685c690f4b2bf8b0e09684a751363529a7

philipp-seitz (approved):

PR #468, opened 2026-05-14
feat: multiple frontend improvements

### LINKED ISSUE #330: [Feat] More info when expanding nodes.
Currently the visualiser displays too little information about nodes.  Could we please add: 
* **The name of the task**.  In the main view, long names are truncated.  Full name should be displayed in the details pane.
* **Inputs and Outputs** the names and types of the inputs and outputs; didn't we have this before?
*  ** Value for inputs and outputs where the are available
* Executor / context for this node.

Image

### LINKED ISSUE #331: [Feat] Visually distinguish nodes of different kinds.
Can we have different visual styles as follow : 
* Inputs and outputs
* Constants
* Everything else 

At the same time, expandable nodes which are not expanded, should look like normal nodes, except for the expansion control.

### LINKED ISSUE #332: [feat] Make key for colours
Put somewhere (sidebar?) a key saying what the colours mean.

### LINKED ISSUE #333: [feat] Tooltips for buttons
In the corner there are mysterious controls.  Add tooltips so the users can figure out what they do.  (The fourth one appears t do nothing, and its icon is incomprehensible.)

Image

### LINKED ISSUE #374: Adjust colours to improve legibility
The colours used for the nodes impede legibility because the contrast between the node fill and the black text is not enough.  Would also recommend changing the background of the not-started nodes to be light grey rather than white (same as background.

Here is a site that allows the check of contrast:
https://rgblind.com/color-blind-color-picker

I would recommend that we target a value of 11 or greater for contrast of chosen node colours against black.  To a first approximation, the following are OK:
green = #aaffaa
red = #ffaaaa
yellow = #ffffaa
grey = #dddddd 

Worth checking that the chosen colours are distinguishable to various forms of colour blindness.  The palette above can be improved in this respect.
https://rgblind.com/color-blind-palette

### LINKED ISSUE #375: Grab handles for expanded nodes
The current dragging behaviour is as follows:
* click-drag a node moves the node (relative to the underlying grid)
* click-drag the background moves the viewport

However when a box is expanded all the area inside the expanded box is treated as the node.  Clicking on what *appears* to be background is actually clicking on the box, and the resulting drag action does not do what is expected.  Worse, even if you know about this behaviour, when dealing with graphs with many (nested) boxes then it can be difficult to impossible to find a bit of background to actually click on.

Oe suggestion is to treat the area drawn as background as if it actually is background, and provide a "handle" on the box for dragging.  

(We might also consider some affordance for resizing the box; not in scope for this issue, but should be considered alongside to avoid making an incompatible design choice.)

### LINKED ISSUE #378: [Feat] Hide nodes inserted by the graphbuilder
Hide nodes that are inserted by the graphbuilder and not by the user:
e.g.:
- fold / unfold when creating maps

philipp-seitz:
This adds small-scale frontend qol improvements:
- [x] adds tooltips to control buttons, closes #333 
- [x] updates the color to the more legible nexus colors, closes #374 
- [x] add a color key, closes #332 
- [x] adds a drag handle and resize button for expanded group nodes, closes #375 
- [x] slightly alters the visuals of constant-, input-, and output nodes, closes #331 
- [x] hide fold/unfold in map nodes, closes #378

~Left to do:~
- [x] add more information to nodes (especially tasks), closes #330 
- [x] Docs for all off the above


philipp-seitz:
I've now added a change to hide fold/unfold nodes added by the builder when inserting map nodes.
The logic is a bit frail, but at the moment, this is the only place we need it. Happy to revert/ or put into a separate pr

tierkreis_visualization/frontend/src/components/colors.ts
```diff
@@ -0,0 +1,25 @@
+export const bg_color = (status: string) => {
+  switch (status) {
+    case "Started":
+      return "bg-nexus-purple text-white";
```
johnchildren: double check with @aidanCQ if we should be using amber here instead

johnchildren (approved):

PR #470, opened 2026-05-18
fix: Small fixes and improvements

### LINKED ISSUE #197: Consolidate noqa: F821 in one place
Currently we write `# noqa: F821 # fmt: skip` at every line it is needed. However we may as well turn it off for a whole file (or even the whole repo) as we use pyright for this kind of check anyway.

### LINKED ISSUE #340: [BUG] Frontend position caching in `tkr-vis-graph`
## Describe the bug
Node positions are not reset after running the command with different graphs

## To Reproduce 
Have graph `a` and `b` in test.py
- Run `tkr-vis-graph test.py:a`
- Run `tkr-vis-graph test.py:b`

### LINKED ISSUE #372: Visualizer should error early when workflow graphs are in an unsupported format
The tierkreis graph format has changed a few times in the past few years and users may have old graphs in their checkpoints directory that are no longer supported. When the visualiser opens these files it can sometimes attempt to display them anyway which leads to strange errors. Instead we should fail early and let the user know their graph is not supported.

### LINKED ISSUE #448: [Bug] Can't run evals on not evaluated path
Running a workflow with an unevaluated if branch that contains an eval or higher-order node can be visualized.
This still works in the ad hoc visualizer and is likely related to `-1` missing

philipp-seitz:
closes: #197 
closes: #448
I added a hint which appears if the tkr version of a graph doesn't match the visualizer, but it still tries to visualize it.
closes: #372
closes: #340


tierkreis/tierkreis/namespace.py
```diff
@@ -132,6 +132,7 @@ def stubs(self) -> str:
 
         return f'''"""Code generated from {self.name} namespace. Please do not edit."""
 
+{NO_QA_STR}
```
johnchildren: Maybe just move this in line now rather that importing it as a variable from another file where it isn't used?

johnchildren (approved):

PR #474, opened 2026-05-27
feat: Legacy graph conversion

johnchildren:
Provides a method to convert the existing graph format into the new internal format used by the rust runtime.


tierkreis/src/graph.rs
```diff
@@ -224,3 +236,462 @@ impl WorkflowGraph {
...
+    pub port_indices: HashMap<NodeIndex, HashMap<String, PortIndex>>,
+    pub port_names: HashMap<PortIndex, String>,
+    pub to_link: HashMap<PortIndex, ValueRef>,
+    pub output_node: Option<NodeIndex>,
+}
+
+impl ConversionState {
+    fn new(total_nodes: usize) -> Self {
+        // Rought heuristic that we will likely have around 2 ports per node.
+        let graph = MultiPortGraph::with_capacity(total_nodes, total_nodes * 2);
```
philipp-seitz, line 327: I think the true value migth be a bit higher?

tierkreis/src/graph.rs
```diff
@@ -224,3 +236,462 @@ impl WorkflowGraph {
...
+
+    fn convert_eval(
+        &mut self,
+        graph_source: (u32, String),
+        inputs: HashMap<String, (u32, String)>,
+        outputs: HashMap<String, Vec<u32>>,
+    ) {
+        let incoming = inputs.len();
+        let outgoing = outputs.len();
+        let node_index = self.graph.add_node(incoming + 1, outgoing);
```
philipp-seitz, line 510: why is this `+1`? we can have evals without an input, or is this the graph

johnchildren: Yes, there is always an additional input from the graph itself

philipp-seitz (dismissed):

philipp-seitz (approved):

PR #483, opened 2026-06-02
feat: SQLite workflowstate

### LINKED ISSUE #481: Add alternative to inmemory workflow state storage
Possible implementations:
- sql
- (File)

philipp-seitz:
closes: #481


tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+    Queryable, Selectable, Identifiable, Insertable, Associations, Debug, Clone, AsChangeset,
+)]
+#[diesel(belongs_to(WorkflowModel, foreign_key = workflow_id))]
+#[diesel(primary_key(id, attempt))]
+#[diesel(table_name = workflow_runs)]
+pub struct WorkflowRunModel {
+    pub id: String,
+    pub attempt: i32,
+    pub workflow_id: String,
+    pub run_metadata: String, // JSON string
```
johnchildren: Seemingly sqlite does support `JSONB` blob storage in more recent versions: https://sqlite.org/draft/jsonb.html

It looks like there is support in diesel for it as well:

https://github.com/diesel-rs/diesel/blob/48b6ec6cd0deeae51f5c6e6f2e953f3f1ac6f6a7/diesel/src/sqlite/types/json.rs

I am a bit unclear on how to write migrations for it though, possibly just setting the column type as `JSONB` might work but the documentation seems slightly conflicting

tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+use std::collections::HashMap;
+use std::hash::BuildHasher;
+
+// -----------------------------------------------------------------------------
+// Workflows
+// -----------------------------------------------------------------------------
+
+#[derive(Queryable, Selectable, Identifiable, Insertable, Debug, Clone, AsChangeset)]
+#[diesel(table_name = workflows)]
+pub struct WorkflowModel {
```
johnchildren: I would just cut the `Workflow` suffix from these struct names, you might be worried about clashing with names from the interface but if that's the case we can just use namespacing instead.

`WorkflowModel` ==> `models::Workflow`

tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+#[diesel(belongs_to(WorkflowModel, foreign_key = workflow_id))]
+#[diesel(primary_key(id, attempt))]
+#[diesel(table_name = workflow_runs)]
+pub struct WorkflowRunModel {
+    pub id: String,
+    pub attempt: i32,
+    pub workflow_id: String,
+    pub run_metadata: String, // JSON string
+    pub status: Option<String>,
+    pub started_at: Option<NaiveDateTime>,
```
johnchildren: You are slightly inconsistent with the timestamp column names, which I guess might be because of the names I used in the node state struct?

If you'd prefer `started_at`, `running_at`, `complete_at` we could change the convention.

tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+use diesel::r2d2::{ConnectionManager, Pool, PooledConnection};
+use miette::miette;
+use std::collections::HashMap;
+use std::hash::BuildHasher;
+
+// -----------------------------------------------------------------------------
+// Workflows
+// -----------------------------------------------------------------------------
+
+#[derive(Queryable, Selectable, Identifiable, Insertable, Debug, Clone, AsChangeset)]
```
johnchildren: I think you probably don't want to derive `AsChangeset` for every struct, only ones where you are doing updates of subsets of the fields. For example `NodeStateUpdate` might have an `AsChangeset` derived implementation which includes the timestamps but not the node_location etc.

tierkreis/src/location.rs
```diff
@@ -65,4 +66,52 @@ impl Location {
...
+        self.0
+            .iter()
+            .map(|component| match component {
+                LocationComponent::Node(index) => index.index(),
+            })
+            .collect()
+    }
+}
+
+impl std::fmt::Display for Location {
```
johnchildren, line 89: https://github.com/Quantinuum/tierkreis/blob/557312340b468ee46532f475093c60b7e51da3dc/tierkreis_core/rust/location.rs

We might be able to re-use the implementations I wrote for Location in this previous PR

tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+/// [`RunAttemptState`] is the full state of a run.
+#[derive(Debug, Default)]
+pub struct RunAttemptState {
+    pub nodes: HashMap<Location, crate::state::interface::NodeState>,
+    pub metadata: HashMap<String, String>,
+}
+
+fn parse_location(raw: &str) -> miette::Result<Location> {
```
johnchildren: You could potentially use the `deserialize_as` attribute for this, see the second example here: https://docs.diesel.rs/master/diesel/deserialize/derive.Queryable.html#examples

tierkreis/migrations/2026-06-01-094031-0000_create_tables/up.sql
```diff
@@ -0,0 +1,43 @@
+-- Your SQL goes here
+CREATE TABLE `node_states`(
+	`id` TEXT NOT NULL PRIMARY KEY,
```
johnchildren: You might want to use `INTEGER PRIMARY KEY` instead here, I'm not sure how a TEXT primary key will perform and INTEGER PRIMARY KEY has the additional benefit that it will auto-increment when you insert new rows.

tierkreis/src/state/models.rs
```diff
@@ -0,0 +1,442 @@
...
+pub struct NodeOutputModel {
+    pub id: String,
+    pub node_state_id: String,
+    pub name: String,
+    pub asset_location: String,
+}
+
+/// [`RunAttemptState`] is the full state of a run.
+#[derive(Debug, Default)]
+pub struct RunAttemptState {
```
johnchildren: I'm not certain what the purpose of this struct is, I don't really want to be collecting the entire state from the database, I want to query it for the state I need?

Let's take some time this afternoon to pair program though this PR as I think there are a lot of improvements we can make to the queries.

tierkreis/src/state/queries.rs
```diff
@@ -0,0 +1,440 @@
...
+use diesel::r2d2::{ConnectionManager, Pool};
+use miette::miette;
+
+use crate::location::Location;
+use crate::state::models::{NodeState, UpsertNodeState, Workflow, WorkflowRun};
+
+/// [`RunAttemptState`] is the full state of a run.
+#[derive(Debug, Default)]
+#[allow(missing_docs)]
+pub struct RunAttemptState {
```
johnchildren: I don't think this is being used anywhere anymore and can be removed right?

tierkreis/src/state/queries.rs
```diff
@@ -0,0 +1,440 @@
...
+    use crate::state::schema::workflows::dsl as wf;
+
+    let mut conn = connection
+        .get()
+        .map_err(|err| miette!("Failed to get SQLite connection from pool: {err}"))?;
+
+    let attempt_i32 = i32::try_from(attempt)
+        .map_err(|_| miette!("Attempt value {attempt} does not fit into i32"))?;
+
+    let workflow_id = uuid::Uuid::nil().to_string();
```
johnchildren, line 102: is this intentionally the nil uuid?

johnchildren (commented):
A couple of small comments and I think the queries could be improved a bit still, but we can probably merge this after clarifying these points.

johnchildren (approved):

PR #491, opened 2026-06-09
Experimental runtime

johnchildren:


.cargo/config.toml
```diff
@@ -11,3 +11,9 @@ rustflags = [
   "-C", "link-arg=-undefined",
   "-C", "link-arg=dynamic_lookup",
 ]
+
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-Clink-arg=-Wl,--no-rosegment"]
```
johnchildren, line 16: This was for profiling, not needed anymore

PR #492, opened 2026-06-10
feat: Implement all remaining nodes for nextgen

johnchildren:
Implements the remaining nodes:
* Implement IfElse
* Implement Loop
* Implement Map

Swap graph to use BTreeMap internally (probably faster with u32 keys)

### DIFF
--- a/tierkreis/src/orchestrator.rs
+++ b/tierkreis/src/orchestrator.rs
@@ -771,7 +771,6 @@ impl Orchestrator {
             .boxed_local())
     }
 
-    #[allow(clippy::too_many_arguments)]
     #[instrument(skip(self, workflow_state, inputs, subgraph), fields(loc = %loc), err)]
     async fn build_subsequent_map_actions(
         &self,
--- a/tierkreis/src/orchestrator.rs
+++ b/tierkreis/src/orchestrator.rs
@@ -4,72 +4,109 @@ and [`AssetStorage`][crate::asset_storage::AssetStorage] implementations to driv
 of [Event]s with updates about each node in the Workflow.
 */
 use std::{
-    collections::HashMap,
-    sync::{Arc, LazyLock, Mutex},
+    collections::{HashMap, HashSet},
+    iter,
+    sync::{Arc, Mutex},
 };
 
+use bitvec::vec::BitVec;
 use futures::{
-    StreamExt,
+    Stream, StreamExt, TryFutureExt,
     channel::mpsc,
-    stream::{BoxStream, select_all},
+    future,
+    stream::{self, BoxStream, LocalBoxStream, select_all},
 };
 use miette::{Context, IntoDiagnostic, miette};
 use portgraph::NodeIndex;
-use serde_json::Value;
+use tracing::instrument;
 
 use crate::{
     asset_storage::{
-        AssetStorageRegistry,
-        interface::{AssetKey, AssetSpec},
-        transfer_assets,
+        AssetStorageRegistry, fold_assets, interface::AssetSpec, load_asset, save_asset,
+        transfer_assets, unfold_asset,
     },
-    event::{Event, send_complete},
+    event::{
+        Event, EventReceiver, EventSender, NodeEvent, send_complete, send_map_elem_complete,
+        send_running_loop, send_running_map, send_running_switching, send_workflow_run_complete,
+    },
     executor::{ExecutorRegistry, interface::TaskPlan},
-    graph::{NodeDefinition, WorkflowGraph},
+    graph::{LegacyWorkflowGraph, NodeDefinition, WorkflowGraph},
     location::Location,
+    state::{WorkflowState, interface::NodeState},
 };
 
-static EMPTY_INPUTS: LazyLock<HashMap<String, AssetSpec>> = LazyLock::new(HashMap::new);
+/// `Action` describes an operation the Orchestrator should perform.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Action {
+    /// The node location in the graph where orchestration should occur.
+    pub loc: Location,
+    /// The kind of action to perform.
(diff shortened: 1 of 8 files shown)


tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
-                    let mut subworkflow_path = context.subworkflow_context.to_vec();
-                    subworkflow_path.push(n);
-                    !context.node_outputs.contains_key(&subworkflow_path)
+                    let (definition, state) = node_states
+                        .get(&n)
+                        .expect("Node definition/state not found");
+
+                    let has_outputs = state.outputs.is_some();
+                    let not_already_scheduled = state.scheduled_time.is_none();
+                    // Always traverse control nodes.
```
philipp-seitz: why is this the case for all of them and not `EagerIfElse`? E.g. once an eval has (all) outputs, the subgraph should be done?

tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
+        let connected_port_name = workflow_graph.get_port_name(connected_port)?;
+
+        let mut outputs = HashMap::new();
+        let value = node_state
+            .outputs
+            .as_ref()
+            .ok_or_else(|| miette!("No outputs found on node connected to `{port_name}` port."))?
+            .get(connected_port_name)
+            .ok_or_else(|| miette!("No outputs found for `{connected_port_name}` port."))?;
+        outputs.insert("value".to_string(), value.clone());
```
philipp-seitz, line 466: As a reminder this keeps our implicit contract of naming ports "value"

tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
+        graph_inputs: Arc<HashMap<String, AssetSpec>>,
+        loc: Location,
+        name: &String,
+    ) -> miette::Result<Action> {
+        let outputs: HashMap<String, AssetSpec> = graph_inputs
+            .iter()
+            .filter(|(k, _)| k == &name)
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+        // TODO: It's a little bit unclear if this is desired behaviour.
```
philipp-seitz: I don't think it is, should discuss

tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
+                loc,
+                kind: ActionKind::SetComplete { outputs },
+            }))
+            .boxed());
+        }
+
+        let stream = self
+            .build_actions(
+                OrchestrationContext {
+                    subworkflow_context: loc,
```
philipp-seitz: Naming is a bit confusing (because it is not a Context), maybe something like "parent"

tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
+        subgraph: Arc<WorkflowGraph>,
+    ) -> miette::Result<LocalBoxStream<'_, miette::Result<Action>>> {
+        let mut input_sets = Vec::new();
+        for mapped_port in mapped_ports {
+            let unfolded_assets =
+                unfold_asset(&self.asset_storage_registry, &inputs, &mapped_port)?;
+
+            if input_sets.is_empty() {
+                input_sets.extend(iter::repeat_n(inputs.clone(), unfolded_assets.len()));
+            }
```
philipp-seitz, line 742: This assumes that all ports will unfold into the same length

tierkreis/src/orchestrator.rs
```diff
@@ -120,184 +164,820 @@ impl Orchestrator {
...
+    #[instrument(skip(self), err)]
+    async fn build_completed_map_action(
+        &self,
+        workflow_graph: Arc<WorkflowGraph>,
+        workflow_state: Arc<dyn WorkflowState>,
+        node_states: Arc<HashMap<NodeIndex, (NodeDefinition, NodeState)>>,
+        mapped_ports: HashSet<String>,
+        n: NodeIndex,
+        loc: Location,
+        completed: BitVec,
```
philipp-seitz: could just pass the length instead

philipp-seitz: also in many cases it seems you want only the output_idx()

tierkreis/src/orchestrator.rs
johnchildren: Good spot, looks like it wasn't needed!

tierkreis/src/orchestrator.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/492/changes/833333195d36272a7c1931f044e96d19d7eadb0c

tierkreis/src/orchestrator.rs
johnchildren: Yeah, I guess we can discuss whether we want to allow multiple ports to be mapped over like this

tierkreis/src/orchestrator.rs
johnchildren: https://github.com/Quantinuum/tierkreis/pull/492/changes/833333195d36272a7c1931f044e96d19d7eadb0c improved here

tierkreis/src/orchestrator.rs
johnchildren: I'll remove and adjust the tests, we can revisit later

johnchildren: https://github.com/Quantinuum/tierkreis/pull/492/changes/9d810e8558df77577171c74d0b66063e688b9f81 done!

philipp-seitz (approved):

PR #493, opened 2026-06-10
chore: Add orchestrator dependencies

johnchildren:
* configure crate-type for python lib and bin targets
* num-complex for handling complex numbers
* `signal` from `tokio` for handling ctrl+c
* `tracing` for debugging and observability

### DIFF
--- a/tierkreis/Cargo.toml
+++ b/tierkreis/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [lib]
+crate-type = ["rlib", "cdylib"]
 
 [[bin]]
 name = "tkr"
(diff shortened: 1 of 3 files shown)


tierkreis/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [lib]
+crate-type = ["rlib", "cdylib"]
```
philipp-seitz, line 7: If we run `maturin develop` outside of uv sync this will produce artifacts in the repository tree at `tierkreis/libtierkreis.dylib` which we might want to exclude in the gitignore

philipp-seitz (approved):

tierkreis/Cargo.toml
johnchildren: Interesting, the path on linux is `tierkreis/tierkreis/_tierkreis.abi3.so` which is already in the gitignore (via .so), so yes that makes sense. Possibly a quick separate PR?

PR #501, opened 2026-06-11
Fix: sql state

### LINKED ISSUE #499: SQLite tests are flakey in CI
Example: https://github.com/Quantinuum/tierkreis/actions/runs/27281179982/attempts/1

I think there might be some race condition in the tests, possibly alleviated by fixtures or by using separate databases for each test.

philipp-seitz:
closes: #499


tierkreis/src/state/queries.rs
```diff
@@ -16,6 +17,16 @@ fn utc_timestamp(ts: NaiveDateTime) -> DateTime<Utc> {
     DateTime::<Utc>::from_naive_utc_and_offset(ts, Utc)
 }
 
+fn encode_map_completed(map_completed: &BitVec) -> miette::Result<Vec<u8>> {
+    serde_json::to_vec(map_completed)
+        .map_err(|err| miette!("Failed to serialize map_completed: {err}"))
```
johnchildren: I have actually been inconsistent about this, but there's actually a miette method for this: https://docs.rs/miette/latest/miette/struct.Error.html#method.wrap_err

It will make a nice chain of errors as well

tierkreis/src/state/sql.rs
```diff
@@ -56,6 +57,38 @@ fn run_migrations(connection: &mut SqliteConnection) -> miette::Result<()> {
...
+    let manager = ConnectionManager::<SqliteConnection>::new(database_url);
+
+    let pool = Pool::builder()
+        .build(manager)
+        .map_err(|_| miette!("Error connecting to {}", database_url))?;
+
+    let mut conn = pool
+        .get()
+        .map_err(|err| miette!("Error acquiring connection for SQLite setup: {err}"))?;
+    if should_run_migrations {
```
johnchildren: I'm not sure if this is necessary, I think `run_pending_migrations` will only run migrations that haven't been run?

johnchildren (approved):

Issue #514, opened 2026-06-17
Nodes are often excessively scheduled multiple times


johnchildren:
As the orchestrator has to emit events for the state of nodes to change, sometimes these changes are not reflected in the state upon the next round of orchestration and nodes are scheduled multiple times, even if an event has already been emitted for them.

While running nodes multiple times is perfectly acceptable, we can mitigate how much this happens with a few measures:

* Allowing the orchestrator to directly write the state of nodes during the `perform_actions` function rather than emitting events.
  * For Const/Input/Output nodes we can directly write that they are complete.
  * For Task nodes we can write that they have been scheduled.
* Allow events to apply to multiple Nodes/Locations, so that we can emit less events during `perform_actions` to reduce the volume of events and therefore avoid performing another scheduling round while we are still writing node states.

johnchildren:
This is somewhat improved now, but still a little shaky (subjectively) and so I think further improvements can be made before we close this issue.

johnchildren:
I think this is resolved with Philipp's changes

PR #515, opened 2026-06-18
feat: Multiple locations in events

johnchildren:
Changes events to cover multiple locations and introduces an `ActionPlan`
struct that can be used as a basis for an independent planning step.

This should reduce the volume of events being produced which should
be a band aid for a potential deadlock when channels fill up, but
further improvements should be made in this area still.

### DIFF
--- a/tierkreis/src/event.rs
+++ b/tierkreis/src/event.rs
@@ -30,7 +30,7 @@ impl Event {
         }
     }
 
-    /// Returns true if the event indicates that a node has some specific outputs.
+    /// Returns the outputs from locations specified in the event if any.
     #[must_use]
     pub fn outputs(self) -> Vec<HashMap<String, AssetSpec>> {
         match self {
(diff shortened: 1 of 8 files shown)


johnchildren:
https://github.com/Quantinuum/tierkreis/issues/514 related

tierkreis/src/event.rs
```diff
@@ -31,10 +32,10 @@ impl Event {
 
     /// Returns true if the event indicates that a node has some specific outputs.
```
philipp-seitz: docs are incorrect now for some of the functions in this file

tierkreis/src/event.rs
```diff
@@ -181,7 +182,7 @@ pub type EventReceiver = mpsc::Receiver<Event>;
 pub async fn send_running(event_sender: &mut EventSender, loc: Location) -> miette::Result<()> {
     event_sender
         .send(Event::Node(NodeEvent {
-            loc,
+            locs: vec![loc],
```
philipp-seitz, line 184: is there a reason we don't clone here?

philipp-seitz (approved):

PR #516, opened 2026-06-19
chore: Improve sqlite query efficiency

johnchildren:
Use a single query for updating node states.

### DIFF
--- a/tierkreis/src/state/queries.rs
+++ b/tierkreis/src/state/queries.rs
@@ -3,12 +3,13 @@ This module defines the queries for reading the workflow state from the `SQlite`
 */
 use std::collections::HashMap;
 use std::hash::BuildHasher;
-use std::ops::BitOr;
 
 use bitvec::vec::BitVec;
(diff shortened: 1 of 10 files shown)


tierkreis/src/state/queries.rs
```diff
@@ -145,342 +134,116 @@ pub fn insert_default_workflowrun(
...
-        cancelled_time: None,
-        error_time: None,
-        cond: None,
-        loop_index: None,
-        map_completed: None,
-        error: None,
-        error_detail: None,
-    };
+    // This is worse than it looks, typically a map_completed event
+    // should only occur for a single location.
```
philipp-seitz: Can we guarantee this? Also in the case we have two MapElementCompleted for the same map, this would generate two rows with different bitvecs, where the later one wins in the update further down

philipp-seitz (changes requested):

philipp-seitz (approved):

PR #521, opened 2026-06-24
feat: Implement AssetKind str format

johnchildren:
Expand AssetKind string format to cover files

### DIFF
--- a/tierkreis/src/asset_storage/interface.rs
+++ b/tierkreis/src/asset_storage/interface.rs
@@ -4,7 +4,8 @@ implementations must satisfy.
 */
 use std::{fmt::Display, path::PathBuf, str::FromStr, time::SystemTime};
 
-use miette::miette;
+use miette::{Context, IntoDiagnostic, miette};
+use url::Url;
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


tierkreis/src/asset_storage/interface.rs
```diff
@@ -35,18 +36,24 @@ impl FromStr for AssetKind {
...
         }
     }
 }
 
 impl Display for AssetKind {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::Memory => write!(f, "memory"),
-            _ => unimplemented!(),
+            Self::Memory => write!(f, "memory://process"),
```
philipp-seitz, line 55: do we extend this in the future?

philipp-seitz (approved):

tierkreis/src/asset_storage/interface.rs
johnchildren: For memory specifically or for other variants? I would expect we will add s3/http Assets and a few more. For memory I am not sure and do not suspect so.

PR #522, opened 2026-06-24
chore: Change complex serialization format

johnchildren:
**Breaking** change the serialization format of complex numbers to the more conventional tuple of real and imaginary parts.

Also changes `get_outputs` to read the output annotation from Workflow graphs to allow direct deserialization into PTypes (including complex numbers)

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -62,7 +62,7 @@ in
       export MATURIN_NO_PROGRESS=1
       export RUST_LOG=error
     '' else ''
-      export LD_LIBRARY_PATH=${pkgs.stdenv.cc.cc.lib}/lib:$LD_LIBRARY_PATH
+      export LD_LIBRARY_PATH=${pkgs.stdenv.cc.cc.lib}/lib:${pkgs.zlib}/lib:$LD_LIBRARY_PATH
       export MATURIN_NO_PROGRESS=1
       export RUST_LOG=error
     '';
--- a/docs/justfile
+++ b/docs/justfile
@@ -1,4 +1,4 @@
-sphinx_opts  := ""
+sphinx_opts  := "-T"
 sphinx_build := "uv run sphinx-build"
 source_dir   := "source"
 build_dir    := "build"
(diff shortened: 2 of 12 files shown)


docs/justfile
```diff
@@ -1,4 +1,4 @@
-sphinx_opts  := ""
+sphinx_opts  := "-T"
```
johnchildren, line 1: This will dump the stack trace if a doc build fails

tierkreis/tests/executor/test_executors.py
```diff
@@ -43,7 +43,7 @@ def test_shell_executor():
     storage.clean_graph_files()
     run_graph(storage, executor, g, {"value": "world"})
     actual_output = read_outputs(g, storage)
-    assert actual_output == b'Hello "world"\n'
+    assert actual_output == "Hello world"
```
johnchildren, line 46: This is a change in behavior for read_outputs, but I think it is closer to the intention

tierkreis/tests/workers/shell_worker/main.sh
```diff
@@ -1,4 +1,4 @@
 #! /usr/bin/env bash
 
-greeting=$(cat $input_greeting_file)
-echo $TEST_FLAG $greeting >> $output_value_file
+greeting=$(cat $input_greeting_file | tr -d '"')
```
johnchildren, line 3: I changed these workers to output json strings rather than bytes to match the type annotations in the graph

devenv.nix
```diff
@@ -62,7 +62,7 @@ in
       export MATURIN_NO_PROGRESS=1
       export RUST_LOG=error
     '' else ''
-      export LD_LIBRARY_PATH=${pkgs.stdenv.cc.cc.lib}/lib:$LD_LIBRARY_PATH
+      export LD_LIBRARY_PATH=${pkgs.stdenv.cc.cc.lib}/lib:${pkgs.zlib}/lib:$LD_LIBRARY_PATH
```
johnchildren, line 65: Required to run the qsci example on my machine seemingly

philipp-seitz (approved):

PR #530, opened 2026-06-25
chore: Rename WorkflowState -> WorkflowRunState

johnchildren:
Better description as to what this state interface is about.

### DIFF
--- a/tierkreis/src/orchestrator.rs
+++ b/tierkreis/src/orchestrator.rs
@@ -32,7 +32,7 @@ use crate::{
     executor::{ExecutorRegistry, interface::TaskPlan},
     graph::{LegacyWorkflowGraph, NodeDefinition, WorkflowGraph},
     location::Location,
-    state::{WorkflowState, interface::NodeState},
+    state::{WorkflowRunState, interface::NodeState},
 };
 
 /// `Action` describes an operation the Orchestrator should perform.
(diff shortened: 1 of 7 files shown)


tierkreis/src/orchestrator.rs
```diff
@@ -103,13 +103,13 @@ struct ActionPlan {
 
 /// The context state for the orchestration
 #[derive(Debug)]
-pub struct OrchestrationContext<WS: WorkflowState> {
+pub struct OrchestrationContext<WS: WorkflowRunState> {
     parent_loc: Location,
     graph_inputs: HashMap<String, AssetSpec>,
     workflow_state: Arc<WS>,
```
philipp-seitz: is it intentional that this field was not change? Also we have some `let workflow_state`

philipp-seitz (approved):

tierkreis/src/orchestrator.rs
johnchildren: ah good point, I'll change those as well!

johnchildren: https://github.com/Quantinuum/tierkreis/pull/530/changes/5145bfe15ca43127bb7785f25a25cc47a953070e addressed here!

philipp-seitz (approved):

PR #532, opened 2026-06-26
fix: uv dependency

philipp-seitz:
Move the uv dependency to have consistent behavior across `uv sync` with `--all-groups`.

### DIFF
--- a/tierkreis/pyproject.toml
+++ b/tierkreis/pyproject.toml
@@ -16,7 +16,7 @@ requires-python = ">=3.12"
 dependencies = [
     "hugr",
     "pydantic~=2.5",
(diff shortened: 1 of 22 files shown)


tierkreis/pyproject.toml
```diff
@@ -16,6 +16,7 @@ requires-python = ">=3.12"
 dependencies = [
     "hugr",
     "pydantic~=2.5",
+    "uv>=0.11.21",
```
johnchildren: this seems like a surprising change, what does this give us?

johnchildren (approved):

johnchildren (approved):

PR #538, opened 2026-07-02
ci: Add wheels release for tierkreis

johnchildren:


tierkreis/src/runtime.rs
```diff
@@ -312,100 +288,3 @@ pub(crate) async fn run_workflow_in_memory<S: BuildHasher>(
 
     Ok(outputs)
 }
-
-/// Run a Workflow in memory from a filepath to a python file containing a builder.
```
johnchildren, line 316: This code isn't really being used and I figured we should just ditch it. If we need it back we can just call `python` as a subprocess.

johnchildren:
I've disabled the workflow for PRs but you can see a successful run here: https://github.com/Quantinuum/tierkreis/actions/runs/28661740664

philipp-seitz (approved):

Issue #512, opened 2026-06-17
`AssetStorage` trait should be async


johnchildren:
Reading/writing files can be done asynchronously with tokio and we want to enable network storage later on, so we should use async methods in the traits rather than blocking. This may impact which functions are async in the orchestrator/executors.

PR #549, opened 2026-07-07
feat: async AssetStorage methods

johnchildren:
Change the methods on the AssetStorage interface to be async and start using the tokio fs implementation. This has some minimal overhead for assets stored in memory but should help with future io based assets including s3 storage etc.


johnchildren:
Resolves: https://github.com/Quantinuum/tierkreis/issues/512

philipp-seitz (approved):

PR #564, opened 2026-07-10
feat: Add config struct for the runtime

johnchildren:


tierkreis/src/runtime.rs
```diff
@@ -14,126 +15,148 @@ use crate::{
...
-        executor_registry.insert(
-            "memory".to_string(),
-            Box::new(InMemoryExecutor::try_new(&asset_storage_registry, "memory").await?),
-        );
-        let executor_registry = Arc::new(executor_registry);
+impl Default for RuntimeConfig {
+    fn default() -> Self {
+        let tierkreis_dir = home_dir()
+            .unwrap_or_else(|| "/tmp".into())
+            .join(".tierkreis");
```
philipp-seitz, line 66: In the old runtime we use a env_var for this, which we hand of to workers as additional context

philipp-seitz (approved):

PR #567, opened 2026-07-13
feat: Allow reading state of multiple nodes

johnchildren:
Also changes `NodeStates` to use `Location` rather than `NodeIndex` to avoid leaking `NodeIndex` into the state interface, which may have some implications later down the line.


tierkreis/src/state/queries.rs
```diff
@@ -372,6 +372,86 @@ pub async fn read_node_state(
...
+                        .map_size
+                        .ok_or_else(|| miette!("Could not get map size from node state"))?
+                        .try_into()
+                        .into_diagnostic()?,
+                );
+                Ok::<_, miette::Report>(bits)
+            })
+            .transpose()?;
+
+        let outputs = read_outputs(conn, &db_node).await?;
```
johnchildren, line 431: In most cases I don't actually need the outputs in the node state for orchestration, so we should probably move this out

johnchildren: It does mean collecting node state as a two step process though.

tierkreis/src/state/queries.rs
```diff
@@ -372,6 +372,86 @@ pub async fn read_node_state(
     }
 }
 
+/// Read the persisted node state for a workflow run at a given location.
```
johnchildren: docstring out of date

philipp-seitz (approved):

PR #576, opened 2026-07-16
feat: rust api

philipp-seitz:
Current routes are missing compared to the old one:
- [ ] /logs
- [ ] /{loc}/logs
- [ ] /errors
- [ ] /metadata
- [x] /{loc}/inputs/portname; I added it but its very messy
- [ ] /restart
Other missing things:
- the websocket
- the openapi stub generation
- viewing graphs during construction
- a solution how to view runs/attempts


tierkreis/src/bin/tkr.rs
```diff
@@ -42,7 +44,17 @@ fn main() -> miette::Result<()> {
             ..
         } => {}
         Command::Serve {} => {
-            tierkreis::server::serve()?;
+            tokio::runtime::Builder::new_multi_thread()
```
johnchildren: The `#[tokio::main]` decorator will also do the same thing as this

tierkreis/src/server/models.rs
```diff
@@ -0,0 +1,256 @@
...
+
+pub type HandlerResult<T> = Result<T, AppError>;
+
+/// Server state shared across all requests.
+#[derive(Clone)]
+pub struct AppState {
+    pub runtime_state: Arc<SqliteRuntimeState>,
+    pub asset_registry: AssetStorageRegistry,
+    pub update_receiver: watch::Receiver<RuntimeWatchState>,
+    pub pool: ConnPool,
```
johnchildren: It's unclear to me why you need a ConnPool here if you have the RuntimeState already?

tierkreis/src/server.rs
```diff
@@ -1,65 +1,103 @@
...
+    }
 
-    let router = router.merge(SwaggerUi::new("/swagger-ui").url("/api-docs/openapi.json", api));
-    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000")
+    let router = router.with_state(app_state);
+    let port = 3000;
+    let listener = tokio::net::TcpListener::bind(("0.0.0.0", port))
         .await
         .into_diagnostic()?;
+    tracing::info!("Visualization server listening on http://localhost:{port}");
```
johnchildren, line 90: We might need to setup a tracing subscriber as I don't think this is currently printing to stdout by default

tierkreis/src/server/nodes.rs
```diff
@@ -0,0 +1,271 @@
...
+) -> miette::Result<serde_json::Value> {
+    let outputs = node_state
+        .outputs
+        .as_ref()
+        .ok_or_else(|| miette::miette!("Node has no outputs"))?;
+    let asset_spec = outputs
+        .get(port_name)
+        .ok_or_else(|| miette::miette!("Missing output port '{port_name}'"))?;
+    let registry = asset_registry.read().await;
+    let bytes = registry
```
johnchildren: You might want to use this utility function here: https://github.com/Quantinuum/tierkreis/blob/727ded7245ecc13a6b892f52bb318814f8ae4d68/tierkreis/src/asset_storage.rs#L60

tierkreis/src/server/nodes.rs
```diff
@@ -0,0 +1,271 @@
...
+/// If the node has no outputs, if any output port is missing, if the storage is not found, or if any value cannot be deserialized.
+pub async fn try_load_outputs(
+    node_state: &NodeState,
+    asset_registry: &AssetStorageRegistry,
+) -> miette::Result<HashMap<String, serde_json::Value>> {
+    let Some(outputs) = &node_state.outputs else {
+        return Ok(HashMap::new());
+    };
+    let mut loaded = HashMap::new();
+    for name in outputs.keys() {
```
johnchildren: Similarly you could use an existing utility function here: https://github.com/Quantinuum/tierkreis/blob/main/tierkreis/src/asset_storage.rs#L35

You would need to map to json for the values afterwards however.

tierkreis/src/server/routes.rs
```diff
@@ -0,0 +1,254 @@
...
+pub async fn list_workflows(
+    State(state): State<AppState>,
+) -> HandlerResult<Json<Vec<WorkflowDisplay>>> {
+    let mut conn = state.pool.get().await.into_diagnostic()?;
+    let summaries = list_workflow_run_summaries(&mut conn).await?;
+
+    let displays: Vec<WorkflowDisplay> = summaries
+        .into_iter()
+        .map(|s| {
+            let id_int = {
```
johnchildren: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.as_u64_pair I would probably just use this method and take the part of the pair you are displaying instead

tierkreis/src/server.rs
```diff
@@ -1,65 +1,103 @@
 /*!
-The server module defines the HTTP interface to the Workflow server.
+The server module defines the HTTP visualization interface to the Workflow server.
```
johnchildren: This might be true now, but I thought the intended scope of this module would include routes beyond what the visualiser uses?

tierkreis/src/server.rs
```diff
@@ -1,65 +1,103 @@
...
+        let index = dist.join("index.html");
+        let assets_dir = dist.join("assets");
+
+        if assets_dir.exists() {
+            use tower_http::services::ServeDir;
+            router = router.nest_service("/assets", ServeDir::new(&assets_dir));
+            tracing::info!("Serving frontend assets from {}", assets_dir.display());
+        }
+        if index.exists() {
+            router = router.fallback(move || {
```
johnchildren: Could you use `ServeFile` here? https://github.com/tokio-rs/axum/blob/b7e37889932edcf521ca54e5ed30245f01180994/examples/static-file-server/src/main.rs#L54

tierkreis/src/state/queries.rs
```diff
@@ -674,3 +674,152 @@ pub async fn read_run_attempt_metadata(
...
+    if db_node.complete_time.is_none() {
+        return Ok(None);
+    }
+
+    let db_outputs: Vec<NodeOutput> = NodeOutput::belonging_to(&db_node)
+        .get_results(conn)
+        .await
+        .into_diagnostic()?;
+
+    let outputs: HashMap<String, AssetSpec> = db_outputs
```
johnchildren: `read_node_state` currently returns the outputs of the node as well: https://github.com/Quantinuum/tierkreis/blob/727ded7245ecc13a6b892f52bb318814f8ae4d68/tierkreis/src/state/queries.rs#L354

So this function might not be necessary?

tierkreis/src/server/routes.rs
```diff
@@ -0,0 +1,254 @@
...
+    // Or the frontend needs to start submitting run_ids
+    let run_state = state
+        .runtime_state
+        .load_workflow_run_state(run_id, 0)
+        .await?;
+    let workflow_id = run_state.workflow_id();
+
+    tracing::info!("Listing nodes for {}", run_id);
+
+    // TODO can we somehow avoid loading the entire graph (e.g. only the nested ones we need)
```
johnchildren, line 102: Not sure entirely what you mean by this TODO?

johnchildren: Is it more about loading the graphs in the loop below?

tierkreis/src/server/routes.rs
```diff
@@ -0,0 +1,254 @@
...
+#[utoipa::path(
+    get,
+    path = "/workflows/",
+    responses((status = OK, body = Vec<WorkflowDisplay>))
+)]
+pub async fn list_workflows(
+    State(state): State<AppState>,
+) -> HandlerResult<Json<Vec<WorkflowDisplay>>> {
+    let mut conn = state.pool.get().await.into_diagnostic()?;
+    let summaries = list_workflow_run_summaries(&mut conn).await?;
```
johnchildren: I think it would be safe move move this into the state interface to avoid needing the connection pool at this level. You can just add an `unimplemented!()` to the InMemoryState implementation if it's awkward to put in there.

tierkreis/src/server/models.rs
johnchildren: It looks like it's for the WorkflowRunSummaries later

tierkreis/src/server/models.rs
```diff
@@ -0,0 +1,256 @@
...
+    Const,
+    /// The graph output node.
+    #[serde(rename = "output")]
+    Output,
+    /// A graph input node.
+    #[serde(rename = "input")]
+    Input,
+}
+
+// TODO can we simply serde NodeDefinition
```
johnchildren: https://docs.rs/serde_plain/latest/serde_plain/ might do the trick?

tierkreis/src/server/nodes.rs
```diff
@@ -0,0 +1,271 @@
...
+                    .ok()
+                    .map(|v| v.to_string()),
+                conditional,
+            });
+        }
+
+        // Determine node value string.
+        let value = match def {
+            NodeDefinition::Const { value } => {
+                Some(serde_json::to_string(value).unwrap_or_default())
```
johnchildren, line 350: I think it may be inadvisable to use `.unwrap_or_default` in a few places here just because it would be more useful to have an error if this fails than an empty field etc.

tierkreis/src/server/nodes.rs
```diff
@@ -0,0 +1,271 @@
...
+
+/// Load a subgraph from a Const node that contains a serialized `WorkflowGraph`.
+fn load_subgraph_from_const_node(
+    workflow_graph: &WorkflowGraph,
+    node_index: portgraph::NodeIndex,
+) -> miette::Result<WorkflowGraph> {
+    tracing::info!("Loading subgraph from Const node {node_index:?}");
+    let Some((source_node, _source_port)) = workflow_graph
+        .connected_input_by_port_name(node_index, "graph")
+        .ok()
```
johnchildren: I think it might be preferable to use `.wrap_err` or similar instead of `.ok` here, `.ok` loses the underlying error context.

johnchildren (commented):
Lots of good progress, I think there are some potential improvements. I think the main thing would be to avoid leaking the Sqlite ConnPool into the server itself, but we can iterate on this design a bit. Ideally it wouldn't be Sqlite specific either.

tierkreis/src/server/routes.rs
philipp-seitz: yeah we basically have to load the entire graph for all nodes no matter the actual nesting level

tierkreis/src/server.rs
```diff
@@ -1,66 +1,112 @@
...
         .nest("/api", api_router)
         .split_for_parts();
+    let mut router =
+        api_http_router.merge(SwaggerUi::new("/swagger-ui").url("/api-docs/openapi.json", api));
+
+    // SPA
+    let dist = std::env::current_dir()
+        .into_diagnostic()
+        .wrap_err("Failed to resolve current working directory")?
+        .join("tierkreis_visualization/tierkreis_visualization/static/dist");
```
philipp-seitz: We need a good solution for the asset dir

tierkreis/src/server.rs
philipp-seitz: I used this for debugging, should I add this or remove the tracing?

johnchildren: Just keep it, we should add a subscriber later.

johnchildren (approved):

PR #578, opened 2026-07-20
feat: Add a Nexus HTTP client

johnchildren:
Pre-requisite work for a Nexus Executor, adds the Nexus endpoints required to upload a hugr, submit a job and get the results.

Possibly controversial as it adds a dependency on hugr, but this could be revised. I figure as we will want to add a hugr dependency later this should be reasonably uncontroversial, but it does mean we would need to make sure the hugr version in the runtime and the hugr we submit using the client match.

### DIFF
--- a/tierkreis/src/executor/nexus/client.rs
+++ b/tierkreis/src/executor/nexus/client.rs
@@ -32,6 +32,12 @@ use crate::executor::nexus::client::models::{
     results::{QSysResult, QSysResultData},
 };
 
+const REFRESH_ENDPOINT: &str = "/auth/tokens/refresh";
+const PROJECTS_ENDPOINT: &str = "/api/projects/v1beta2";
+const HUGR_ENDPOINT: &str = "/api/hugr/v1beta";
+const JOBS_ENDPOINT: &str = "/api/jobs/v1beta3";
+const QSYS_RESULTS_PARTIAL_ENDPOINT: &str = "/api/qsys_results/v1beta2/partial";
+
 #[derive(Deserialize)]
 struct AccessToken {
     data: AccessTokenData,
@@ -164,6 +170,7 @@ pub struct NexusClient {
     http1_client: Client,
     client: Client,
     base_url: Url,
+    base_ws_url: Url,
 }
 
 impl NexusClient {
@@ -207,6 +214,9 @@ impl NexusClient {
             serde_json::from_str(&refresh_token_contents).into_diagnostic()?;
 
         let base_url: Url = format!("{scheme}://{host}").parse().into_diagnostic()?;
+        let mut base_ws_url: Url = base_url.clone();
+        base_ws_url.set_scheme("wss");
+
         let jar = Jar::default();
         jar.add_cookie_str(
             &format!("myqos_oat={}", refresh_token.data.refresh_token),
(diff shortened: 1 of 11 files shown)


tierkreis/src/executor/nexus/client.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let mut refresh_token_contents = String::new();
+        refresh_token_file
+            .read_to_string(&mut refresh_token_contents)
+            .await
+            .into_diagnostic()?;
+        let refresh_token: RefreshToken =
+            serde_json::from_str(&refresh_token_contents).into_diagnostic()?;
+
+        let base_url: Url = format!("http://{host}").parse().into_diagnostic()?;
```
johnchildren: We might need to make the scheme configurable as I was having trouble getting mockito to work with TLS but we should be using TLS by default (and I don't think we should ever _not_ be using it).

tierkreis/src/executor/nexus/client/models/jobs.rs
```diff
@@ -0,0 +1,196 @@
...
+    n_shots: u64,
+    n_qubits: u64,
+}
+
+impl NewExecuteJobItem {
+    pub fn new(program_id: Uuid, n_shots: u64) -> Self {
+        NewExecuteJobItem {
+            program_id,
+            n_shots,
+            n_qubits: 20,
```
johnchildren, line 46: We will want to make this configurable 😆

tierkreis/src/executor/nexus/client/models/hugr.rs
```diff
@@ -0,0 +1,66 @@
...
+impl NewHugrRelationships {
+    pub fn new(project_id: Uuid) -> Self {
+        Self {
+            project: NewRelationship::new(project_id, "project"),
+        }
+    }
+}
+
+#[derive(Serialize)]
+struct HugrProperties {}
```
philipp-seitz, line 22: I assume this is left empty intentionally

tierkreis/src/executor/nexus/client/models/jobs.rs
philipp-seitz: can the exectuor modify the job? This would/could be part of the resource annotation?

johnchildren: I don't think this can change once submitted, but yes this would be ideal for the resource definition rather than the task parameters I think.

tierkreis/src/executor/nexus/client/models/hugr.rs
johnchildren: Yes, we need a json object there to satisfy the API I believe. I would like to populate it later with some pre-defined things like the location of the node and the run_id later but this will be easier with a small change in Nexus that I am having reviewed.

tierkreis/src/executor/nexus/client.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        response.error_for_status_ref().into_diagnostic()?;
+
+        Ok(())
+    }
+
+    pub async fn find_project_data(&self, name: &str) -> miette::Result<Option<Data>> {
+        let url = self
+            .base_url
+            .join("/api/projects/v1beta2")
```
philipp-seitz: maybe we can have a list of apis we use at the top of the file? Than at least we know where to check?

philipp-seitz (approved):

tierkreis/src/executor/nexus/client.rs
johnchildren: Not sure if it's what you meant but I've moved the endpoints into constants at the top of: `tierkreis/src/executor/nexus/client.rs`

philipp-seitz (approved):

PR #579, opened 2026-07-20
feat: Add a Nexus executor

johnchildren:
Adds a nexus executor that has a bundled in worker for submitting, monitoring and collecting the results of a single hugr job

### DIFF
--- a/tierkreis/src/executor/nexus.rs
+++ b/tierkreis/src/executor/nexus.rs
@@ -511,7 +511,7 @@ mod tests {
             .await;
             send_message(
                 &mut socket,
-                &json!({"status": "RUNNING", "message": "job is runing"}),
+                &json!({"status": "RUNNING", "message": "job is running"}),
             )
             .await;
             send_message(
--- a/tierkreis/src/executor/nexus.rs
+++ b/tierkreis/src/executor/nexus.rs
@@ -268,7 +268,7 @@ impl NexusExecutor {
         asset_storage_registry: &AssetStorageRegistry,
         output_storage_name: &str,
     ) -> miette::Result<Self> {
-        let client = NexusClient::try_new(client::Scheme::Https, host, None).await?;
+        let client = NexusClient::try_new(client::TLSMode::Default, host, None).await?;
 
         let asset_storage_registry_lock = asset_storage_registry.read().await;
         if !asset_storage_registry_lock.contains_key(output_storage_name) {
--- a/tierkreis/src/executor/nexus.rs
+++ b/tierkreis/src/executor/nexus.rs
@@ -13,18 +13,17 @@ use std::{
 use futures::{
     FutureExt, SinkExt, StreamExt, TryStreamExt,
     channel::mpsc,
-    future::BoxFuture,
+    future::{self, BoxFuture},
     stream::{BoxStream, FuturesUnordered},
 };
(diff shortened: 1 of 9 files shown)


tierkreis/src/executor/nexus.rs
```diff
@@ -1,6 +1,1088 @@
...
+                    | StatusEnum::Depleted
+                    | StatusEnum::Cancelled => break 'outer,
+                }
+            }
+
+            // Reconnect if the job status was not a final state.
+            job_status_stream = client
+                .listen_for_job_status(internal_task.job_id)
+                .await
+                .wrap_err("Failed to listen for job status")?;
```
philipp-seitz, line 261: In which cases does this happen?

tierkreis/src/executor/nexus.rs
```diff
@@ -1,6 +1,1088 @@
...
+            },
+            result
+                .map_err(|err| miette!("Failed to join task: {err}"))
+                .flatten(),
+        )
+    });
+
+    let key = (workflow_run_id, attempt, loc.clone());
+    // If we are meant to cancel this job, signal this to the cancellation routine.
+    if let Some(InternalJobMonitoringStatus::Cancelling) = job_handles.get(&key) {
```
philipp-seitz, line 287: Should we implement this for other executors?

tierkreis/src/executor/nexus.rs
```diff
@@ -1,6 +1,1088 @@
...
+        client_config: &NexusClientConfig,
+        asset_storage_registry: &AssetStorageRegistry,
+        output_storage_name: &str,
+    ) -> miette::Result<Self> {
+        let client = NexusClient::try_new(client_config).await?;
+
+        let asset_storage_registry_lock = asset_storage_registry.read().await;
+        if !asset_storage_registry_lock.contains_key(output_storage_name) {
+            return Err(miette!("output_storage_name not in registry"));
+        }
```
philipp-seitz, line 408: Might be worthwhile to think about a nexus storage

tierkreis/src/executor/nexus.rs
```diff
@@ -1,6 +1,1088 @@
...
+
+    fn execute(&self, task_plans: Vec<TaskPlan>) -> BoxFuture<'_, miette::Result<()>> {
+        async move {
+            let mut task_sender = self.task_sender.clone();
+            self.client.refresh_tokens().await?;
+
+            for task_plan in task_plans {
+                if task_plan.worker_name != "nexus_worker" {
+                    return Err(miette!("Unknown worker: `{}`", task_plan.worker_name));
+                }
```
philipp-seitz, line 510: This fuses worker+executor, I'm not sure where the boundaries here should be

philipp-seitz (approved):

tierkreis/src/executor/nexus.rs
johnchildren: When the server disconnects early

tierkreis/src/executor/nexus.rs
johnchildren: Yeah we probably should, or at least handle this edge case consistently

tierkreis/src/executor/nexus.rs
johnchildren: I agree, I think it will simplify the hugr/circuit upload as well

PR #587, opened 2026-08-05
feat: Add hugr i64_s and f64 operations

johnchildren:
Adds many new builtin tasks for numbers. Not a complete list but a decent start.

### DIFF
--- a/tierkreis/src/executor/inmemory.rs
+++ b/tierkreis/src/executor/inmemory.rs
@@ -170,7 +170,9 @@ fn run_builtin(
         "itostring_s" => un_method!(inputs, outputs, to_string, i64),
 
         // f64 operations
+        #[allow(clippy::float_cmp)]
         "feq" => bin_op!(inputs, outputs, ==, f64),
+        #[allow(clippy::float_cmp)]
         "fne" => bin_op!(inputs, outputs, !=, f64),
         "flt" => bin_op!(inputs, outputs, <, f64),
         // igt is the legacy name
@@ -185,8 +187,8 @@ fn run_builtin(
         "fabs" => un_method!(inputs, outputs, abs, f64),
         "fmul" => bin_op!(inputs, outputs, *, f64),
         "idiv" => bin_op!(inputs, outputs, /, f64),
-        "ffloor" => bin_op!(inputs, outputs, /, f64),
-        "fceil" => bin_op!(inputs, outputs, /, f64),
+        "ffloor" => un_method!(inputs, outputs, floor, f64),
+        "fceil" => un_method!(inputs, outputs, ceil, f64),
         "ftostring" => un_method!(inputs, outputs, to_string, f64),
 
         // "Value" operations.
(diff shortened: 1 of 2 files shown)


tierkreis/src/executor/inmemory.rs
```diff
@@ -104,60 +104,97 @@ fn output_value(
...
+}
+
+fn run_builtin(
+    task_name: &str,
+    inputs: &HashMap<String, Vec<u8>>,
+) -> miette::Result<HashMap<String, Vec<u8>>> {
+    let mut outputs = HashMap::new();
+    match task_name {
+        // i64 operations
+        "ieq" => bin_op!(inputs, outputs, ==, i64),
```
acl-cqc, line 148: I had been hypothesizing a single polymorphic method `eq<A>(A,A) -> bool`  that I was going to map just to `eq`; not sure how useful it is to have a separate int-only version. (How's it different - just that it errors if the args aren't int?)

What about return type? I guess we don't need to specify, the type is just for internal conversion of arguments (and fail if the args are wrong)?

tierkreis/src/executor/inmemory.rs
```diff
@@ -104,60 +104,97 @@ fn output_value(
...
+        "ilt_s" => bin_op!(inputs, outputs, <, i64),
+        // igt is the legacy name
+        "igt_s" | "igt" => bin_op!(inputs, outputs, >, i64),
+        "ilte_s" => bin_op!(inputs, outputs, <=, i64),
+        "igte_s" => bin_op!(inputs, outputs, >=, i64),
+        "imax_s" => bin_method!(inputs, outputs, max, i64),
+        "imin_s" => bin_method!(inputs, outputs, min, i64),
+        "iadd" => bin_op!(inputs, outputs, +, i64),
+        "isub" => bin_op!(inputs, outputs, -, i64),
+        "ineg" => un_op!(inputs, outputs, -, i64),
```
acl-cqc, line 159: purely presentational but might be nice to put unary before binary (so `//i64 unary operators` then `//i64 binary operators` then `//f64 unary...` and so on)

acl-cqc (approved):
I *haven't* been through checking for completeness (and I can see that e.g. conversions are missing), but looks good to me

tierkreis/src/executor/inmemory.rs
johnchildren: Makes sense, I put them in the order they appear in the hugr spec but I can change it if you would prefer that ordering.

tierkreis/src/executor/inmemory.rs
johnchildren: I think the return type is currently being inferred by the rust compiler from the implementation of `==`.

There is a "dynamic" `eq` further down in the match statement that should work for all json serializable values, I guess the advantage of this is parsing could be a bit quicker and as a check in case you do accidentally pass something that isn't an integer?

Mostly I just copied the hugr spec without thinking about it too much however.

PR #590, opened 2026-08-07
feat:central logging

### LINKED ISSUE #571: Implement centralised logging for the new runtime

philipp-seitz:
closes: #571


tierkreis/src/logging.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn default() -> Self {
+        let tierkreis_log = home_dir()
+            .unwrap_or_else(|| "/tmp".into())
+            .join(".tierkreis/tierkreis.log");
+
+        LoggingConfig {
+            log_file: Some(tierkreis_log),
+            log_format: LogFormat::Compact,
+            log_level: Some("info".to_string()),
+            otel_endpoint: Some("http://localhost:4317".to_string()),
```
johnchildren: I think we should probably default this to "None" out of the box.

tierkreis/src/executor/subprocess.rs
```diff
@@ -384,7 +397,18 @@ fn spawn_worker(
     worker_name: &str,
     worker_args_path: &Path,
 ) -> miette::Result<tokio::process::Child> {
+    let mut headers = HashMap::new();
+    let cx = tracing::Span::current().context();
+    opentelemetry::global::get_text_map_propagator(|p| p.inject_context(&cx, &mut headers));
```
johnchildren: I feel like there might be a better way to do this environment variable injection.

Looking at the otel spec it seems there is some specific documentation about how to do this with some reference implementations: https://opentelemetry.io/docs/specs/otel/context/env-carriers/

[Python implementation](https://github.com/open-telemetry/opentelemetry-python/blob/a6d59ca6304ad8d9e419d6a4179437eb5c99590c/opentelemetry-api/src/opentelemetry/propagators/_envcarrier.py#L73).

Annoyingly I don't see something out of the box in the rust sdk, but I think it should be possible to make a struct that wraps a reference to `Command` and uses the [env](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.env) method to inject normalised versions.

tierkreis/src/executor/subprocess.rs
```diff
@@ -20,10 +20,12 @@ use regex::Regex;
 use serde::{Deserialize, Serialize};
 use tempfile::NamedTempFile;
 use tokio::{
-    io::AsyncReadExt,
+    io::{AsyncBufReadExt, BufReader},
     process::Command,
     task::{AbortHandle, JoinHandle},
 };
+use tracing::Instrument as _;
```
johnchildren: what is importing as `_` doing here?

tierkreis/tierkreis/worker/worker.py
```diff
@@ -171,33 +180,48 @@ def run(self, worker_definition_path: Path) -> None:
         node_definition = self.storage.read_call_args(worker_definition_path)
         logger.debug(node_definition.model_dump())
 
+        carrier = {
```
johnchildren, line 184: Similarly there are utilities in the python sdk for this extraction: https://github.com/open-telemetry/opentelemetry-python/blob/a6d59ca6304ad8d9e419d6a4179437eb5c99590c/opentelemetry-api/src/opentelemetry/propagators/_envcarrier.py#L35

tierkreis/tierkreis/worker/worker.py
```diff
@@ -216,8 +240,26 @@ def app(self, argv: list[str]) -> None:
                 "either provide--stubs-path or the path to the node definition"
             )
             return
+
+        otel_endpoint = os.getenv(
+            "OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4317"
```
johnchildren: similarly we probably should default to trying not to export by default unless this env var is set.

tierkreis/tierkreis/worker/worker.py
```diff
@@ -216,8 +240,26 @@ def app(self, argv: list[str]) -> None:
...
+        )
+        set_global_textmap(TraceContextTextMapPropagator())
+        otel_provider = TracerProvider(
+            resource=Resource({
+                "service.name": f"tierkreis::worker::{self.name}",
```
johnchildren: I've not seen this pattern for names before in otel, was this from an example or similar?

I would expect something more like

"service.name": self.name,
"service.namespace": "tierkreis_workers",

tierkreis/src/executor/subprocess.rs
```diff
@@ -398,12 +422,21 @@ fn spawn_worker(
...
+) -> tokio::task::JoinHandle<String> {
     let mut stderr = child.stderr.take().unwrap();
     tokio::spawn(async move {
+        let mut reader = BufReader::new(&mut stderr).lines();
         let mut stderr_out = String::new();
-        let _ = stderr.read_to_string(&mut stderr_out).await;
-
+        while let Ok(Some(line)) = reader.next_line().await {
+            // TODO: parse for actual log level, python log will be stderr by default.
+            tracing::info!(target: "tierkreis::worker", run_id = %run_id, loc = %loc, "{line}");
```
johnchildren, line 451: If the workers are also exporting logs themselves does it make sense to be reading and relogging them here as well? I think this could get quite unreadable in i.e. map nodes.

tierkreis/src/monitoring.rs
```diff
@@ -0,0 +1,225 @@
+/*!
+This module defines the central logging capabilities of the runtime.
```
johnchildren, line 2: This module is doing a lot more than logging now as it contains code for tracing/metrics/logs. I would possibly rename it to monitoring.

tierkreis/src/logging.rs
```diff
@@ -0,0 +1,225 @@
...
+    sync::OnceLock,
+};
+use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+use tracing_opentelemetry::{MetricsLayer, OpenTelemetryLayer};
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::fmt;
+use tracing_subscriber::fmt::writer::BoxMakeWriter;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::util::SubscriberInitExt as _;
```
johnchildren: Why is this being imported as `_` ?

tierkreis/src/logging.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+/// Initialize the runtime subscriber with logging and OpenTelemetry.
+pub fn init_logging_and_tracing(logging_config: Option<LoggingConfig>) {
+    init(&logging_config.unwrap_or_default(), true);
+}
+
+/// Flush and shut down the global tracer provider.
+pub fn flush_tracing() {
+    if let Some(provider) = TRACER_PROVIDER.get() {
```
johnchildren: As far as I can tell you only create `TRACE_PROVIDER` so you can use it here. But it would also be available from the `global` module of opentelemetry here: https://docs.rs/opentelemetry/0.32.0/opentelemetry/global/fn.tracer_provider.html

tierkreis/src/runtime.rs
```diff
@@ -202,6 +211,25 @@ impl Runtime {
                     let workflow_state = state
                         .load_workflow_run_state(workflow_run_id, attempt)
                         .await?;
+                    let workflow_id = workflow_state.workflow_id().to_string();
+                    match event.clone() {
```
johnchildren: This clone might not be necessary if you match on `&event` I think.

tierkreis/tierkreis/worker/worker.py
```diff
@@ -171,33 +180,48 @@ def run(self, worker_definition_path: Path) -> None:
...
-        try:
-            function = self.functions.get(node_definition.function_name, None)
-            if function is None:
-                msg = (
-                    f"{self.name}: function name\n"
-                    f"{node_definition.function_name} not found"
+        with tracer.start_as_current_span(
+            node_definition.function_name,
+            context=ctx,
+            attributes={"worker.name": self.name},
```
johnchildren: For consistency we could do something like `tierkreis.worker.name` for keys like this.

tierkreis/migrations/2026-06-01-094031-0000_create_tables/up.sql
```diff
@@ -64,4 +64,18 @@ CREATE TABLE `workflow_run_inputs` (
...
-)
+);
+
+CREATE TABLE `executor_debug` (
+	`id` INTEGER NOT NULL PRIMARY KEY,
+	`node_state_id` INTEGER NOT NULL,
+	`executor_name` TEXT NOT NULL,
+	`worker_name` TEXT NOT NULL,
+	`task_name` TEXT NOT NULL,
+	`resources` BLOB NOT NULL CHECK (json_valid(resources, 8)) DEFAULT (jsonb('{}')),
```
johnchildren, line 75: I am a bit unclear about whether this is the resources that were requested for the task or the resources that were allocated for the task

tierkreis/migrations/2026-06-01-094031-0000_create_tables/up.sql
```diff
@@ -64,4 +64,18 @@ CREATE TABLE `workflow_run_inputs` (
 	`storage_name` TEXT NOT NULL,
 	`asset_key` TEXT NOT NULL,
 	FOREIGN KEY (`workflow_run_id`) REFERENCES `workflow_runs` (`id`)
-)
+);
+
+CREATE TABLE `executor_debug` (
```
johnchildren, line 69: I am a bit uncertain about whether this should be a separate table to the node_states table, I can see the reason for doing so might be because these columns will be null for some nodes, but I am not a huge fan of this table being populated by the orchestrator rather than from events from executor itself.

Having said that, I guess the executor isn't aware of it's own name so that might also not work very well.

tierkreis/migrations/2026-06-01-094031-0000_create_tables/up.sql
```diff
@@ -64,4 +64,18 @@ CREATE TABLE `workflow_run_inputs` (
...
+	`id` INTEGER NOT NULL PRIMARY KEY,
+	`node_state_id` INTEGER NOT NULL,
+	`executor_name` TEXT NOT NULL,
+	`worker_name` TEXT NOT NULL,
+	`task_name` TEXT NOT NULL,
+	`resources` BLOB NOT NULL CHECK (json_valid(resources, 8)) DEFAULT (jsonb('{}')),
+	`environment` BLOB NOT NULL CHECK (json_valid(environment, 8)) DEFAULT (jsonb('{}')),
+	`internal_id` TEXT DEFAULT NULL,
```
johnchildren, line 77: What is the internal_id here? The handle from the executor?

johnchildren: It looks like this is populated by a new variant of running event, I think this should certainly be on the node_states table.

tierkreis/src/monitoring.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+fn make_writer(path: Option<&Path>) -> BoxMakeWriter {
+    if let Some(path) = path {
+        let dir = path.parent().unwrap_or_else(|| Path::new("."));
+        let file = path.file_name().unwrap_or_else(|| "tierkreis.log".as_ref());
+        let appender = tracing_appender::rolling::never(dir, file);
+        let (non_blocking, guard): (NonBlocking, WorkerGuard) =
+            tracing_appender::non_blocking(appender);
+        LOG_GUARD.set(guard).expect("log guard already set");
```
johnchildren, line 171: I don't really understand what this LOG_GUARD is meant to be doing?

tierkreis/src/runtime.rs
```diff
@@ -202,6 +211,57 @@ impl Runtime {
...
+                    let workflow_id = workflow_state.workflow_id().to_string();
+                    if let WorkflowRunEvent::NodeEvent(NodeEvent {
+                        locs,
+                        status:
+                            crate::event::NodeStatus::Running {
+                                state_update: Some(RunningStateUpdate::Executor { internal_id }),
+                            },
+                    }) = &event
+                    {
+                        for loc in locs {
```
johnchildren, line 223: As a general principle, if you find yourself doing these kind of state updates in a loop, please change the function to set the state from a list (in this case a list of locations).

tierkreis/src/orchestrator.rs
```diff
@@ -1102,6 +1122,28 @@ impl Orchestrator {
...
+        if !plan.tasks.is_empty() {
+            tracing::info!(
+                "Dispatching {} tasks to executor '{}'",
+                plan.tasks.len(),
+                default_executor_name
+            );
+        }
+        for task in &plan.tasks {
+            workflow_run_state
+                .write_executor_debug_data(ExecutorDebugInformation {
```
johnchildren, line 1134: What we are effectively doing is saving the task plans right? Maybe we should refactor the table to represent that instead?

PR #591, opened 2026-08-13
Feat/restore state

### LINKED ISSUE #484: Make workflow execution more resilient
For the new workflow server, we want the folllowing:
- [ ] Restore the workflow / runtime state after a crash / pkill
- [ ] Reconnect to the executors and check for scheduled tasks (e.g. find the job id of a slurm job)
- [ ] Restore the DB / save the state

philipp-seitz:
closes: #484


tierkreis/src/executor/interface.rs
```diff
@@ -86,4 +89,12 @@ pub trait Executor: Send + Sync {
         attempt: u32,
         task_locations: Vec<Location>,
     ) -> BoxFuture<'_, miette::Result<()>>;
+
+    /// Poll the current status of tasks that are currently known to the executor.
+    ///
+    /// By default you should use `NodeStatus::Unknown`
+    fn known_tasks(
+        &self,
+        tasks: Vec<(Uuid, u32, Location)>,
```
johnchildren: I've used this pattern in a few places now, I wonder if it's worth having a type alias or struct for `(Uuid, u32, Location)`

tierkreis/src/event.rs
```diff
@@ -157,6 +161,8 @@ pub enum NodeStatus {
     Running {
         /// An update to the state of the Node to apply.
         state_update: Option<RunningStateUpdate>,
+        /// Unique handle to the task on the executor that is running this node.
+        handle: Option<TaskHandle>,
```
johnchildren: I think this probably shouldn't be in the `Running` state, or at least we may want to have a new state between scheduled and queued to represent a task that has been accepted by the API/WMS/Kernel and has been assigned a handle but hasn't started yet?

tierkreis/src/executor/interface.rs
```diff
@@ -45,6 +51,38 @@ pub struct TaskPlan {
...
+    Nexus {
+        /// The job identifier assigned by Nexus.
+        job_id: String,
+    },
+}
+
+impl TaskHandle {
+    /// Return the executor registry name responsible for this task.
+    #[must_use]
+    pub fn executor_name(&self) -> &'static str {
```
johnchildren: My expectation for executor names was that they wouldn't be hardcoded like this but rather configured by the user.

tierkreis/src/executor/interface.rs
```diff
@@ -45,6 +51,38 @@ pub struct TaskPlan {
...
+pub enum TaskHandle {
+    /// A task running as a subprocess.
+    Subprocess {
+        /// The process identifier.
+        pid: u32,
+        /// The process start time in Linux clock ticks.
+        start_time: u64,
+    },
+    /// A task running in-memory.
+    InMemory,
```
johnchildren: Maybe having this variant is more self-explanatory, but I would have expected that tasks like this would be represented by the handle being `None` which implies that it needs to start from the beginning (which is true for all executors where the handle is `None`).

tierkreis/src/executor/inmemory.rs
```diff
@@ -254,16 +256,19 @@ fn run_builtin(
 }
 
 type RunningFutures = FuturesUnordered<JoinHandle<BackgroundTask>>;
-type AbortHandles = HashMap<(Uuid, u32, Location), AbortHandle>;
+type AbortHandles = Arc<Mutex<HashMap<(Uuid, u32, Location), AbortHandle>>>;
```
johnchildren: I am not sure what the purpose of this change is? I don't see anything other than the background task mutating this hashmap?

tierkreis/src/state/sql.rs
```diff
@@ -335,6 +353,21 @@ impl RuntimeState for SqliteRuntimeState {
         }
         .boxed()
     }
+
+    fn restore_active_runs(&self) -> BoxFuture<'_, miette::Result<()>> {
+        async move {
+            let _lock = self.lock.read().await;
+            let mut conn = self.get_conn().await?;
+            let interrupted = list_active_runs(&mut conn).await?;
+            for (run_id, attempt) in interrupted {
```
johnchildren: I think you might be able to do this loop inside the closure in `send_modify` to reduce the amount of locking/updates on the watch channel.

tierkreis/src/state/interface.rs
```diff
@@ -99,6 +102,14 @@ pub trait RuntimeState: Debug + Send + Sync {
...
     fn list_workflow_run_summaries(&self)
     -> BoxFuture<'_, miette::Result<Vec<WorkflowRunSummary>>>;
+
+    /// Restore active runs from storage.
+    ///
+    /// Must be called once on startup after a potential crash to re-populate the
+    /// in-memory [`RuntimeWatchState`] from any runs that were not in a terminal
+    /// state when the process last exited. Implementations without durable storage
+    /// may treat this as a no-op.
+    fn restore_active_runs(&self) -> BoxFuture<'_, miette::Result<()>>;
```
johnchildren: Is there ever a case where we _don't_ want to do this on startup or want to do it mid-run? I wonder if maybe this should just be the responsibility of the workflow state object when it is initialized?

tierkreis/src/executor/subprocess.rs
```diff
@@ -541,6 +591,33 @@ impl Executor for SubprocessExecutor {
...
+                        if process_identity(pid).is_ok_and(|(_, current_start_time)| {
+                            current_start_time == start_time
+                        }) =>
+                    {
+                        NodeStatus::Running {
+                            state_update: None,
+                            handle: Some(TaskHandle::Subprocess { pid, start_time }),
+                        }
+                    }
+                    _ => NodeStatus::Unknown,
```
johnchildren: This seems like it should result in an error to me rather than an unknown status?

tierkreis/src/state/sql.rs
```diff
@@ -386,17 +419,44 @@ impl WorkflowRunState for SqliteWorkflowRunState {
...
+                WorkflowRunEvent::Errored {} => {
+                    workflow_update.error_time = Some(now);
+                    send_workflow_stopped = true;
+                }
+                WorkflowRunEvent::Completed {} => {
+                    workflow_update.complete_time = Some(now);
+                    send_workflow_stopped = true;
+                }
                 WorkflowRunEvent::NodeEvent(ref node_event) => {
+                    // TODO: can we receive node events before start is set?
```
johnchildren, line 449: theoretically yes, is that likely to be an issue? Ideally it shouldn't happen but I don't know if it's necessarily an error, just an anomaly?

tierkreis/src/state/schema.rs
```diff
@@ -27,6 +27,7 @@ diesel::table! {
         loop_index -> Nullable<Integer>,
         map_size -> Nullable<Integer>,
         map_completed -> Nullable<Binary>,
+        handle -> Nullable<Text>,
```
johnchildren, line 30: I think this PR is mostly using json to serialize these handles now, but we store them as text rather than as jsonb binary like the other json columns we have. Is there a reason for this? I would guess possibly readability but maybe this was an oversight?

tierkreis/src/executor/subprocess.rs
```diff
@@ -79,16 +79,19 @@ type CancelSender = mpsc::Sender<(Uuid, u32, Location)>;
 type CancelReceiver = mpsc::Receiver<(Uuid, u32, Location)>;
 
 type RunningFutures = FuturesUnordered<JoinHandle<BackgroundTask>>;
-type AbortHandles = HashMap<(Uuid, u32, Location), AbortHandle>;
+type AbortHandles = Arc<Mutex<HashMap<(Uuid, u32, Location), AbortHandle>>>;
```
johnchildren: I don't see a reason why this needs to be put in a mutex anymore either?

tierkreis/src/runtime.rs
```diff
@@ -169,6 +184,117 @@ impl Runtime {
...
+    /// 1. Restores the set of active runs from durable state.
+    /// 2. For each interrupted run, polls the executors to discover whether any
+    ///    tasks are still running, have completed, or were lost.
+    /// 3. Records any newly discovered completions in state.
+    /// 4. Continues the run attempts
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the state backend, executor, or event writing fails.
+    async fn recover(&mut self) -> miette::Result<()> {
```
johnchildren: I'm not very convinced by this function, it seems to duplicate a lot of what the runtime does already.

The orchestrator already uses the node state to make scheduling decisions and as that now includes a handle to the process, would it not make sense to just include that in the arguments we pass to the `execute` method with the stipulation that the executor re-uses processes it is already aware of?

tierkreis/src/executor/subprocess.rs
johnchildren: Oh right I didn't see the guard on the previous match statement, this is meant to cover both the cases when the handle is for the wrong kind of executor and also when we can't query the pid, the process id doesn't exist anymore etc?

johnchildren (commented):
I think there are some good ideas here, but I think some of these changes are a bit more complicated than they need to be, specifically we probably don't need dedicated routines for restoring state if we just pass existing task_ids to the `execute` method as part of normal orchestration tor restore state.

tierkreis/src/executor/nexus.rs
```diff
@@ -507,6 +517,15 @@ impl NexusExecutor {
 
         Ok(job_data.id())
     }
+
+    /// Checks whether a job referenced is already on nexus
+    /// This is a simple check and does not guarantee anything about the
+    /// state of the job.
+    async fn is_job_active(&self, job_id: Uuid) -> miette::Result<Uuid> {
+        self.client.refresh_tokens().await?;
```
johnchildren: refreshing tokens happens at the start of the function, so hopefully we don't need to do this again here, unless you are expecting latency when there are multiple tasks?

tierkreis/src/executor/nexus.rs
```diff
@@ -551,28 +570,43 @@ impl Executor for NexusExecutor {
...
-                let hugr_id = self
-                    .upload_hugr(project_data.id(), &job_name, hugr_package)
-                    .await?;
-
-                let job_id = self
-                    .start_single_job(project_data.id(), &job_name, hugr_id, n_shots)
-                    .await?;
+                // If we were given a handle to a previously submitted Nexus job
+                // and it's still active, reattach to it instead of resubmitting.
+                let restored_job_id = match &task_plan.task_handle {
```
johnchildren: I would probably use the `if let Some(...) = task_plan.task_handle` pattern here

tierkreis/src/executor/interface.rs
```diff
@@ -43,6 +49,29 @@ pub struct TaskPlan {
...
+    Subprocess {
+        /// The process identifier.
+        pid: u32,
+        /// The process start time in Linux clock ticks.
+        start_time: u64,
+    },
+    /// A task running on nexus.
+    Nexus {
+        /// The job identifier assigned by Nexus.
+        job_id: String,
```
johnchildren: Considering you parse the Uuid explicitly in the Executor you may as well make the type `Uuid` here to avoid duplicated parsing.

tierkreis/src/executor/interface.rs
```diff
@@ -43,6 +49,29 @@ pub struct TaskPlan {
...
+    /// An executor-specific identifier for a task that can be persisted and restored.
+    /// Obligation of the [Executor] to check this when a task is received.
+    pub task_handle: Option<TaskHandle>,
+}
+
+/// An executor-specific identifier for a task that can be persisted and restored.
+/// None indicates Inmemory/needs to be scheduled
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "executor", content = "task")]
+pub enum TaskHandle {
```
johnchildren: I am not entirely convinced that having a dedicated struct with all variants is the right thing to do here actually, what about cases where someone is using Tierkreis as a library and wants to add their own new executor? I think maybe we should use `task_handle: Option<String>` in task plan and have every executor decide how they want to parse the task handle.

tierkreis/src/executor/nexus.rs
```diff
@@ -551,28 +570,43 @@ impl Executor for NexusExecutor {
...
+                    Some(TaskHandle::Nexus { job_id }) => {
+                        let job_id = Uuid::parse_str(job_id)
+                            .into_diagnostic()
+                            .wrap_err("Invalid job_id in restored task handle")?;
+                        Some(self.is_job_active(job_id).await?)
+                    }
+                    _ => None,
+                };
+
+                let job_id = if let Some(job_id) = restored_job_id {
```
johnchildren: You could alternatively just merge the two blocks together here.

tierkreis/src/executor/subprocess.rs
```diff
@@ -782,25 +879,35 @@ mod tests {
...
-            workflow_run_id: Uuid::nil(),
-            attempt: 0,
-            event: WorkflowRunEvent::NodeEvent(NodeEvent {
-                locs: vec![loc2.clone()],
-                status: NodeStatus::Running { state_update: None }
-            })
-        }));
+        assert_eq!(events.len(), 6);
+        for loc in [loc1.clone(), loc2.clone()] {
+            assert!(events.iter().any(|event| matches!(
```
johnchildren: I think you might still be able to use .contains still here? The nested loop seems slightly more complex than it needs to be.

tierkreis/src/orchestrator.rs
```diff
@@ -149,6 +155,11 @@ pub struct Orchestrator {
 
     default_storage_name: String,
     asset_storage_registry: AssetStorageRegistry,
+
+    // (run_id, attempt, location) tuples for nodes that have been scheduled
+    // by this Orchestrator instance. This solves relying on scheduled_time == Some()
+    // after a runtime restart.
+    scheduled_this_lifetime: Mutex<HashSet<(Uuid, u32, Location)>>,
```
johnchildren: I think it might be better to use a type like `RWLock<HashMap<(Uuid, u32), Mutex<HashSet<Location>>>`, otherwise this precludes us from running multiple workflows concurrently (even though we don't yet).

johnchildren (commented):
Some preliminary comments, I will follow up with more tomorrow.

tierkreis/src/executor/inmemory.rs
johnchildren: Bumping this comment as I don't think any of the changes in this file are necessary?

tierkreis/src/executor/nexus.rs
```diff
@@ -219,11 +219,21 @@ async fn monitor_task(
     let workflow_run_id = internal_task.workflow_run_id;
     let attempt = internal_task.attempt;
     let job_id = internal_task.job_id;
+    let mut queued_event_sender = event_sender.clone();
+    send_queued(
```
johnchildren, line 223: Would it make more sense to emit this event after `self.start_single_job` in `execute`? I'm not sure if we want another Queued event in the case where we are reconnecting to an existing job?

tierkreis/src/executor/subprocess.rs
```diff
@@ -172,17 +179,39 @@ async fn process_finished_task(
...
+
+    // If the task has a handle, check if the process is still running and terminate it if so.
+    // Other executors should reattach if possible.
+    if let Some(TaskHandle::Subprocess { pid, start_time }) = internal_task.handle {
+        let is_original_process = match process_identity(pid) {
+            Ok(identity) => identity == (pid, start_time),
+            Err(_) => false,
+        };
+        if is_original_process {
+            let status = Command::new("kill")
```
johnchildren, line 199: I guess the justification as to why we kill the process here is because we don't know where the inputs and outputs of the process will live after it terminates? It could be worth adding a comment about that if it's something we plan to change later.

tierkreis/src/monitoring.rs
```diff
@@ -213,16 +213,18 @@ fn init(config: &LoggingConfig, with_telemetry: bool) {
                     .with(fmt_layer)
                     .with(tracing)
                     .with(metrics);
-                tracing::subscriber::set_global_default(subscriber)
-                    .expect("Setting default subscriber failed");
+                if let Err(err) = tracing::subscriber::set_global_default(subscriber) {
```
johnchildren, line 216: If we can't set the tracing subscriber I'm not sure if the line below will work? Maybe we need to use `eprintln` or similar?

tierkreis/src/orchestrator.rs
johnchildren: `RWLock<HashMap<(Uuid, u32), **Arc**<Mutex<HashSet<Location>>>>`

I realized last night you would likely need an Arc as well. As this data is owned by the Orchestrator you may also not need the outer RWLock but I am not certain.

tierkreis/src/runtime.rs
```diff
@@ -129,7 +132,7 @@ enum ExecutorConfig {
 #[derive(Deserialize)]
 enum RuntimeStateConfig {
     Memory {},
-    Sqlite { memory: bool },
+    Sqlite { memory: bool, url: Option<String> },
```
johnchildren, line 141: I think `memory` being true and having a url are mutually exclusive right? We might want to revisit this config later.

tierkreis/src/orchestrator.rs
```diff
@@ -216,10 +230,33 @@ impl Orchestrator {
...
+                &workflow_graph,
+                &node_states,
+                run_id,
+                attempt,
+                &scheduled_this_lifetime,
+            )
+            .collect()
+        };
         // Mark all ready nodes as scheduled.
         Self::mark_nodes_scheduled(&context, ready_nodes.iter()).await?;
```
johnchildren: It will be interesting to see if we don't need this anymore

tierkreis/src/orchestrator.rs
```diff
@@ -341,23 +381,28 @@ impl Orchestrator {
...
                         .node_definition(n)
                         .expect("Node definition not found");
 
                     if let Some(state) = node_states.get(&parent_location.with_node(n)) {
-                        // TODO: This sometimes means even const/input nodes
-                        // are run multiple times if their state is yet
-                        // to be updated from the last orchestration round.
+                        // A node scheduled in a previous process will be rescheduled
+                        // The executor deals with reattaching
+                        let scheduled_this_lifetime = scheduled_this_lifetime.contains(&(
```
johnchildren: `.contains` feels like it could be expensive here when we have a lot of nodes. I wonder if it would be worth combining the node_states and scheduled_this_lifetime data structures so we only need to do one lookup by location.

tierkreis/src/orchestrator.rs
johnchildren: I'm also wondering if this should live in the `OrchestrationContext` rather than the Orchestrator, the reason being that if we want to clean up runs that have completed/error'd, there is an easy enough way to do this in the Runtime (by looking at active_runs and pruning anything not active from this data structure).

tierkreis/src/executor/nexus.rs
philipp-seitz: We don't have access to the correct even_sender there I think?
The queued event currently means "accepted by the executor" so emmiting it twice would be correct?

johnchildren (approved):

PR #595, opened 2026-08-17
chore: instrumentation

philipp-seitz:
Adds instrumentation to (python) workers and executors

### DIFF
--- a/tierkreis/src/executor/inmemory.rs
+++ b/tierkreis/src/executor/inmemory.rs
@@ -10,12 +10,12 @@ use std::{
 };
 
 use futures::{
+    FutureExt, SinkExt, StreamExt,
     channel::mpsc,
     future::BoxFuture,
     stream::{BoxStream, FuturesUnordered},
-    FutureExt, SinkExt, StreamExt,
 };
-use miette::{miette, IntoDiagnostic};
+use miette::{IntoDiagnostic, miette};
 use num_complex::Complex64;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -23,10 +23,10 @@ use tokio::task::{AbortHandle, JoinHandle};
 use uuid::Uuid;
 
 use crate::{
-    asset_storage::{load_assets, save_assets, AssetStorageRegistry},
+    asset_storage::{AssetStorageRegistry, load_assets, save_assets},
     event::{
-        send_cancelled, send_complete, send_error, send_running, EventReceiver, EventSender,
-        RuntimeEvent,
+        EventReceiver, EventSender, RuntimeEvent, send_cancelled, send_complete, send_error,
+        send_running,
     },
     executor::interface::{Executor, TaskPlan, WorkerSpec},
     location::Location,
@@ -559,9 +559,11 @@ mod tests {
         let workers = executor.workers().await?;
 
         // We should expect the builtin worker to be available.
-        assert!(workers
-            .iter()
-            .any(|workers| workers.worker_name == "builtin"));
+        assert!(
+            workers
+                .iter()
+                .any(|workers| workers.worker_name == "builtin")
+        );
 
         Ok(())
     }
--- a/tierkreis/src/executor/inmemory.rs
+++ b/tierkreis/src/executor/inmemory.rs
@@ -10,12 +10,12 @@ use std::{
 };
 
 use futures::{
-    FutureExt, SinkExt, StreamExt,
     channel::mpsc,
     future::BoxFuture,
     stream::{BoxStream, FuturesUnordered},
+    FutureExt, SinkExt, StreamExt,
 };
-use miette::{IntoDiagnostic, miette};
+use miette::{miette, IntoDiagnostic};
 use num_complex::Complex64;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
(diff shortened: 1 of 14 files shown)


tierkreis/src/monitoring.rs
```diff
@@ -65,15 +65,23 @@ impl Default for LoggingConfig {
...
+/// pending spans.
 pub fn flush_logs() {
     if let Ok(mut guard) = LOG_GUARD.lock() {
         drop(guard.take());
     }
+    // Shutdown the tracer provider to ensure all spans are flushed
+    if let Ok(mut provider) = TRACER_PROVIDER.lock()
+        && let Some(p) = provider.take()
+    {
+        let _ = p.shutdown();
```
johnchildren, line 83: https://docs.rs/opentelemetry/latest/opentelemetry/global/fn.tracer_provider.html I think you should be able to just use this instead

tierkreis/src/executor/inmemory.rs
```diff
@@ -377,7 +390,7 @@ async fn process_tasks(
                     Ok(ok) => ok,
                     Err(err) => panic!("Failed to join to future: {err}"),
                 };
-
+                tracing::debug!( workflow_run_id = %background_task.workflow_run_id, attempt = %background_task.attempt, loc = %background_task.loc, "Task completed");
```
johnchildren: Please add some newlines here

tierkreis/src/monitoring.rs
```diff
@@ -65,15 +65,23 @@ impl Default for LoggingConfig {
 
 static LOG_GUARD: Mutex<Option<WorkerGuard>> = Mutex::new(None);
 static LOGGING_INITIALIZED: OnceLock<()> = OnceLock::new();
+static TRACER_PROVIDER: Mutex<Option<SdkTracerProvider>> = Mutex::new(None);
```
johnchildren, line 68: See my other comment on line 83, but I think you can just use the `opentelemetry::global` module for this purpose.

tierkreis/src/executor/nexus.rs
```diff
@@ -217,56 +222,62 @@ async fn monitor_task(
...
-                    | StatusEnum::Depleted
-                    | StatusEnum::Cancelled => break 'outer,
                 }
-            }
 
-            // Reconnect if the job status was not a final state.
-            job_status_stream = client
-                .listen_for_job_status(internal_task.job_id)
+                // Reconnect if the job status was not a final state.
+                tracing::info!("Reconnecting to job status stream");
```
johnchildren: I think this would make more sense at `debug` level, this can happen if the web server restarts or similar and isn't usually worth remarking on.

tierkreis/src/executor/nexus.rs
```diff
@@ -217,56 +222,62 @@ async fn monitor_task(
...
+                job_status_stream = client
+                    .listen_for_job_status(internal_task.job_id)
+                    .await
+                    .wrap_err("Failed to listen for job status")?;
+            }
+            tracing::info!("Job status stream ended, processing finished task");
```
johnchildren: Similarly I think this could be debug level

tierkreis/src/executor/nexus.rs
```diff
@@ -217,56 +222,62 @@ async fn monitor_task(
 
     let task_loc = loc.clone();
     let background_loc = loc.clone();
-    let outputs = internal_task.outputs;
+    let outputs = internal_task.outputs.clone();
```
johnchildren: why does this need to be cloned now? I don't see `outputs` being used for anything new in this file.

tierkreis/src/executor/subprocess.rs
```diff
@@ -380,11 +395,24 @@ impl SubprocessExecutor {
     }
 }
 
+struct CommandEnvCarrier<'a>(&'a mut Command);
+impl opentelemetry::propagation::Injector for CommandEnvCarrier<'_> {
+    fn set(&mut self, key: &str, value: String) {
+        self.0.env(key.to_uppercase(), value);
```
johnchildren: You may also want to include a `.replace("-", "_")` on the keys here as well to future proof this

tierkreis/src/executor/subprocess.rs
```diff
@@ -393,17 +421,25 @@ fn spawn_worker(
...
+) -> tokio::task::JoinHandle<String> {
     let mut stderr = child.stderr.take().unwrap();
     tokio::spawn(async move {
+        let mut reader = BufReader::new(&mut stderr).lines();
         let mut stderr_out = String::new();
-        let _ = stderr.read_to_string(&mut stderr_out).await;
-
+        while let Ok(Some(line)) = reader.next_line().await {
+            // TODO: parse for actual log level, python log will be stderr by default.
+            tracing::info!(target: "tierkreis::worker", run_id = %run_id, loc = %loc, "{line}");
```
johnchildren: I'm still reasonably unsure if this is something we actually want to do, I can see this becoming a bit too verbose to be usable. If we are adding an opentelemetry sdk to the worker sdk, is there a more structured approach we can take to logging?

I can see an argument for Executors to capture the logs from whatever they are running and dumping them somewhere but this will start logging everything output by subprocesses at info level from the main process.

tierkreis/tierkreis/worker/worker.py
```diff
@@ -170,34 +180,48 @@ def run(self, worker_definition_path: Path) -> None:
         """
         node_definition = self.storage.read_call_args(worker_definition_path)
         logger.debug(node_definition.model_dump())
+        getter = EnvironmentGetter()
+        carrier = {
+            "traceparent": getter.get(os.environ, "traceparent"),
+            "tracestate": getter.get(os.environ, "tracestate"),
+        }
+        ctx = propagate.extract(carrier)
```
johnchildren: why not just just `propagate.extract(os.environ)` ?

tierkreis/tierkreis/worker/worker.py
```diff
@@ -216,8 +240,42 @@ def app(self, argv: list[str]) -> None:
...
+        if otel_endpoint is None:
+            logger.warning("OTEL_EXPORTER_OTLP_ENDPOINT not set, tracing disabled")
+            return None
+
+        try:
+            set_global_textmap(TraceContextTextMapPropagator())
+            otel_provider = TracerProvider(
+                resource=Resource({
+                    "service.name": self.name,
+                    "service.namespace": "tierkreis_workers",
```
johnchildren:
```suggestion
                    "service.namespace": "tierkreis.workers",
```

johnchildren (approved):

PR #606, opened 2026-08-19
chore: Delay initial websocket ping

johnchildren:
This seems to improve a flakiness issue with the websocket during the cancellation test and should improve the quality of websocket error messages.

Basically the issue was that the ping would result in packets that were unread when the socket was closed, causing an error.

Not a full solution as we need to handle the close handshake a bit better.

Suggested by gpt-5.6-sol.

### DIFF
--- a/tierkreis/src/executor/nexus/client.rs
+++ b/tierkreis/src/executor/nexus/client.rs
@@ -13,7 +13,7 @@ use miette::{Context, IntoDiagnostic, miette};
 use reqwest::{Client, ClientBuilder, cookie::Jar};
 use reqwest_websocket::{Bytes, Message, Upgrade};
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


tierkreis/src/executor/nexus/client.rs
```diff
@@ -114,7 +115,7 @@ impl Stream for JobStatusStream {
                 Poll::Ready(None)
             }
             Poll::Ready(Some(Err(err))) => {
-                Poll::Ready(Some(Err(miette!("Websocket error: {err}"))))
+                Poll::Ready(Some(Err(miette!("Websocket error: {err:?}"))))
```
johnchildren, line 118: Apparently the `Debug` instance of `reqwest-websocket::Error` is more informative but it's a bit ugly.

philipp-seitz (approved):

PR #610, opened 2026-08-20
Improve executor test coverage

johnchildren:


tierkreis/src/executor/subprocess.rs
```diff
@@ -950,14 +1095,27 @@ mod tests {
...
+            RuntimeEvent::WorkflowRun {
+                event: WorkflowRunEvent::NodeEvent(NodeEvent {
+                    status: NodeStatus::Error {
+                        detail: Some(ref detail),
+                        ..
+                    },
+                    ..
+                }),
+                ..
+            } if !detail.is_empty()
```
johnchildren: This probably could just be part of the same assert matches above.

PR #617, opened 2026-08-24
feat: Vibe coded NexusProjectAssetStorage

johnchildren:
Just using this as reference for a suggestion on how to implement this feature.

### DIFF
--- a/tierkreis/src/asset_storage/interface.rs
+++ b/tierkreis/src/asset_storage/interface.rs
@@ -31,6 +31,11 @@ pub enum AssetKind {
         /// in the filesystem.
         root: PathBuf,
     },
+    /// An Asset represented by a resource in a Quantinuum Nexus project.
+    NexusProject {
+        /// The Nexus project containing the resource.
+        project_id: Uuid,
+    },
 }
 
 impl FromStr for AssetKind {
@@ -45,6 +50,15 @@ impl FromStr for AssetKind {
                     miette!("Failed to parse root folder location, for url: {url}")
                 })?,
             }),
+            "nexus" => {
+                let project_id = url
+                    .path_segments()
+                    .and_then(Iterator::last)
+                    .ok_or_else(|| miette!("Missing Nexus project id in url: {url}"))?
+                    .parse()
+                    .into_diagnostic()?;
+                Ok(Self::NexusProject { project_id })
+            }
             scheme => Err(miette!("Unknown scheme: {scheme}")),
         }
     }
@@ -55,8 +69,21 @@ impl Display for AssetKind {
         match self {
             Self::Memory => write!(f, "memory://process"),
             Self::File { root } => write!(f, "file://{}", root.to_string_lossy()),
+            Self::NexusProject { project_id } => write!(f, "nexus://project/{project_id}"),
         }
     }
+}
+
+/// A semantic hint for storage implementations that map byte payloads to typed resources.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum AssetDataType {
+    /// Uninterpreted bytes.
+    #[default]
+    Opaque,
+    /// A binary HUGR package envelope.
+    Hugr,
+    /// A JSON-encoded pytket circuit.
+    Circuit,
 }
 
 /// [`AssetSpec`] describes how an Asset should be stored.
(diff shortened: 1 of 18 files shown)


tierkreis/src/asset_storage/interface.rs
```diff
@@ -55,10 +69,23 @@ impl Display for AssetKind {
         match self {
             Self::Memory => write!(f, "memory://process"),
             Self::File { root } => write!(f, "file://{}", root.to_string_lossy()),
+            Self::NexusProject { project_id } => write!(f, "nexus://project/{project_id}"),
```
johnchildren, line 72: I guess it's weird that this isn't just a HTTP url to the Nexus project itself

tierkreis/src/asset_storage/interface.rs
```diff
@@ -55,10 +69,23 @@ impl Display for AssetKind {
...
         }
     }
 }
 
+/// A semantic hint for storage implementations that map byte payloads to typed resources.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum AssetDataType {
```
johnchildren, line 79: I think this is similar to what we discussed in the team about how to provide type hints

tierkreis/src/runtime.rs
```diff
@@ -414,6 +425,7 @@ impl Runtime {
 
 async fn executor_registry_from_config(
     asset_storage_registry: &AssetStorageRegistry,
+    nexus_storages: &HashMap<String, NexusProjectAssetStorage>,
```
johnchildren, line 428: This seems like a quite bad code smell

tierkreis/src/runtime.rs
```diff
@@ -121,8 +127,7 @@ enum ExecutorConfig {
         output_storage_name: String,
     },
     Nexus {
-        client_config: NexusClientConfig,
```
johnchildren, line 124: This is kind of interesting, I guess it is re-using the clients from the new storage class instead of maintain its own in this setup.

tierkreis/src/executor/nexus.rs
```diff
@@ -104,6 +105,7 @@ async fn process_cancelled_task(
 #[instrument(skip_all, err)]
 async fn process_finished_task(
     client: &NexusClient,
+    nexus_storage: &NexusProjectAssetStorage,
```
johnchildren, line 108: This is quite stinky when `asset_storage_registry` is also being passed around.

tierkreis/src/asset_storage/interface.rs
philipp-seitz: I don't know if we need a new scheme for everything. Would there be a point of having a  more generic storage (e.g. "cloud") which is identified by an HTTP uri?

tierkreis/src/asset_storage/interface.rs
philipp-seitz: Yeah I feel like this might become necessary at one point. Could also be useful for compressed data e.g. we have a gzip type or soemthing

tierkreis/src/asset_storage/nexus.rs
```diff
@@ -0,0 +1,324 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("NexusProjectAssetStorage")
+            .field("storage_name", &self.storage_name)
+            .field("project_name", &self.project_name)
+            .field("project_id", &self.project_id)
+            .finish_non_exhaustive()
+    }
+}
+
+impl NexusProjectAssetStorage {
```
philipp-seitz, line 72: A thing that I was thinking about when writing the config files. This would always be a user config level right? And I think it should not be responsible for creating the project

tierkreis/src/asset_storage/interface.rs
johnchildren: Yeah I think that could be reasonable, I looked around a bit afterwards and spark has something like "s3" as a protocol for resources stored there. I think "nexus" resources could be described partially as a ["jsonapi" resources](https://jsonapi.org/), but exactly how we represent that is slightly non-obvious to me.

johnchildren: I guess you would still need something like "upload_hugr" as a workflow task in order to create a jsonapi resource out of the bytes payload and a name etc.

tierkreis/src/asset_storage/interface.rs
johnchildren: Yeah I think trying out some alternative serialization formats like msgpack so we aren't as tied to json could be helpful as well.

tierkreis/src/asset_storage/nexus.rs
johnchildren: I think if we go down the root of jsonapi resources, we might be able to just provide the project resource as an input to creating the jsonapi resource actually, I'll try and throw something together today/tomorrow.

PR #626, opened 2026-08-28
feat: Add a JSON:API AssetKind

johnchildren:
Adds a JsonAPI variant to AssetKind for supporting JSON:API resources.

Should enable support of resources hosted in Nexus as well as JSON:API compliant APIs.


tierkreis/src/asset_storage/interface.rs
```diff
@@ -45,6 +53,16 @@ impl FromStr for AssetKind {
                     miette!("Failed to parse root folder location, for url: {url}")
                 })?,
             }),
+            "http+jsonapi" => {
```
philipp-seitz, line 56: Not sure if this is the best scheme for JSON:API. They use http scheme with a mime type `application/vnd.api+json`.

PR #635, opened 2026-09-04
feat: Vibe coded python interface

johnchildren:
Not really happy with this still but worth taking a look at to draw some ideas.

### DIFF
--- a/tierkreis/src/asset_storage/file.rs
+++ b/tierkreis/src/asset_storage/file.rs
@@ -58,6 +58,12 @@ impl AssetStorage for FileAssetStorage {
     fn save(&self, key: &AssetKey, value: Vec<u8>) -> BoxFuture<'_, miette::Result<()>> {
         let location = self.location(key);
         async move {
+            tokio::fs::create_dir_all(&self.base_dir)
+                .await
+                .into_diagnostic()
+                .wrap_err_with(|| {
+                    miette!("Could not create asset directory: {:?}", self.base_dir)
+                })?;
             let mut file = File::create(&location)
                 .await
                 .into_diagnostic()
--- a/tierkreis/src/lib.rs
+++ b/tierkreis/src/lib.rs
@@ -20,15 +20,21 @@ pub mod state;
 #[pyo3::pymodule]
 #[pyo3(name = "_tierkreis")]
 mod tierkreis {
-    use std::collections::HashMap;
+    use std::{collections::HashMap, path::PathBuf, sync::Arc};
 
-    use miette::{Diagnostic, IntoDiagnostic};
+    use miette::{Diagnostic, IntoDiagnostic, miette};
     use num_complex::Complex64;
     use pyo3::{FromPyObject, PyErr, Python, exceptions::PyValueError, prelude::*, types::PyBytes};
     use serde::{Deserialize, Serialize};
     use tracing::info;
+    use uuid::Uuid;
 
-    use crate::{graph::LegacyWorkflowGraph, runtime};
+    use crate::{
+        asset_storage::{AssetStorageRegistry, load_assets, save_assets},
+        graph::{LegacyWorkflowGraph, WorkflowGraph},
+        runtime::{self, Runtime as NativeRuntime, RuntimeConfig},
+        state::{RuntimeState, interface::NodeState},
+    };
 
     #[allow(clippy::unnecessary_wraps)]
     #[pymodule_init]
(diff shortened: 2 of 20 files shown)


tierkreis/src/state/inmemory.rs
```diff
@@ -50,7 +50,7 @@ struct RunAttemptState {
 /// references.
 #[derive(Debug, Default)]
 struct InMemoryRuntimeStateInner {
-    workflows: DashMap<Uuid, WorkflowGraph>,
+    workflows: DashMap<Uuid, (Option<String>, WorkflowGraph)>,
```
johnchildren, line 53: We should certainly do this, I think not having it was an oversight.

tierkreis/src/state/interface.rs
```diff
@@ -82,8 +82,12 @@ pub struct NodeState {
 pub trait RuntimeState: Debug + Send + Sync {
     /// Retrieve the [`WorkflowGraph`] specified by id.
     fn load_workflow(&self, workflow_id: Uuid) -> BoxFuture<'_, miette::Result<WorkflowGraph>>;
-    /// Save a [`WorkflowGraph`] and return a new id.
-    fn save_workflow(&self, workflow_graph: WorkflowGraph) -> BoxFuture<'_, miette::Result<Uuid>>;
+    /// Save a named [`WorkflowGraph`] and return a new id.
+    fn save_workflow(
+        &self,
+        name: Option<String>,
```
johnchildren, line 88: Another obvious change

tierkreis/tests/controller/test_error.py
```diff
@@ -36,40 +34,62 @@ def non_zero_exit_code() -> Workflow[EmptyModel, TKR[int]]:
...
-        run_graph(storage, executor, g.data, {}, n_iterations=1000)
-    assert (storage.logs_path.parent / "-/_error").exists()
+@pytest.fixture
+def runtime(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
+    """A subprocess runtime with the failing test worker on PATH."""
+    executable = tmp_path / "tkr-failing-worker"
+    executable.write_text(
+        f"#!{sys.executable}\n"
+        "import runpy\n"
+        f"runpy.run_path({str(WORKER_PATH / 'failing_worker/main.py')!r}, run_name='__main__')\n"
```
johnchildren, line 44: Not sure what's going on here

johnchildren: I guess it's a hack to get around some of the test workers not being in PATH

tierkreis/src/lib.rs
```diff
@@ -104,70 +109,497 @@ mod tierkreis {
...
+        name: Option<String>,
+        status: String,
+        started_time: Option<String>,
+        metadata: HashMap<String, String>,
+        nodes: HashMap<String, NodeStateSnapshot>,
+    }
+
+    #[pyclass(name = "Runtime")]
+    struct PyRuntime {
+        tokio: tokio::runtime::Runtime,
```
johnchildren, line 232: I think there might be simpler ways to do this kind of thing with pyo3's newer features, but I'm not completely certain.

There are some convenience features in the pyo3_async_runtimes crate that might be preferable if we prefer a global runtime: https://docs.rs/pyo3-async-runtimes/latest/pyo3_async_runtimes/tokio/fn.get_runtime.html

tierkreis/src/asset_storage/file.rs
```diff
@@ -39,43 +36,43 @@ impl FileAssetStorage {
...
         }
         .boxed()
     }
 
-    fn save(&self, key: &AssetKey, value: Vec<u8>) -> BoxFuture<'_, miette::Result<AssetKind>> {
+    fn save(&self, key: &AssetKey, value: Vec<u8>) -> BoxFuture<'_, miette::Result<()>> {
         let location = self.location(key);
         async move {
-            let mut file = File::create(&location)
+            tokio::fs::create_dir_all(&self.base_dir)
```
johnchildren, line 61: Not really sure why the model did this, presumably in cases where the location doesn't exist? I think this should probably be handled in the constructor instead (and possibly be an error instead if the user specifies a directory that doesn't exist).

tierkreis/tests/controller/test_runtime_handle.py
```diff
@@ -0,0 +1,56 @@
+from pathlib import Path
+
+import pytest
+
+from tests.controller.sample_graphdata import simple_ifelse
+from tierkreis import Runtime
+
+
+def test_uploaded_workflow_can_back_multiple_runs() -> None:
```
johnchildren, line 9: This is probably a useful test to keep

tierkreis/tierkreis/_tierkreis.pyi
```diff
@@ -0,0 +1,68 @@
+from pathlib import Path
```
johnchildren, line 1: It would be good to include generated python bindings

PR #637, opened 2026-09-04
feat: new hpc executors

philipp-seitz:
Scope:
- Generic HPC Executors with script generation
- Job Spec as Resources in tasks (Kubernetes like)
- Config Update


tierkreis/src/executor/hpc/slurm.j2
```diff
@@ -0,0 +1,43 @@
...
+{% endfor %}
+# --- Environment Setup ---
+{% for name, value in job.environment|items %}export {{ name }}={{ value }}
+{% endfor %}
+#SBATCH --export=ALL
+# --- Load Modules ---
+{% for module in job.modules %}module load {{ module }}
+{% endfor %}
+# --- User Command ---
+{% if job.mpi %}eval mpiexec {{ job.command }}
```
johnchildren: Do you still need the `eval` here? In person you mentioned that this was for psi/j compatibility

tierkreis/src/executor/hpc/slurm.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parses_slurm_job_statuses() {
+        let statuses = parse_job_statuses(
+            b"1|PENDING|0:0|\n2|RUNNING|0:0|\n3|COMPLETED|0:0|\n4|CANCELLED by 42|0:15|\n5|TIMEOUT|1:0|\n",
```
johnchildren: This test might be a bit easier to read if you use multiline strings: https://stackoverflow.com/questions/29483365/what-is-the-syntax-for-a-multiline-string-literal

tierkreis/src/executor/hpc/spec.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Container settings.
+    pub container: Option<ContainerSpec>,
+    /// Opaque command executed by the scheduler.
+    pub command: String,
+    /// Environment exported by the job script.
+    pub environment: HashMap<String, String>,
+    /// Modules loaded before the command.
+    pub modules: Vec<String>,
+    /// Explicit scheduler output path.
+    pub output_path: Option<PathBuf>,
```
johnchildren, line 40: This is the output path for the log file of stdout?

tierkreis/src/executor/hpc/spec.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Explicit scheduler error path.
+    pub error_path: Option<PathBuf>,
```
johnchildren, line 42: This is the output path for the log file of stderr?

tierkreis/src/executor/hpc/spec.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Number of compute nodes requested.
+    pub nodes: u32,
+    /// Number of CPU cores requested on each node.
+    pub cores_per_node: Option<u32>,
+    /// Memory requested on each node, in gibibytes.
+    pub memory_per_node_gb: Option<u32>,
+    /// Number of GPUs requested on each node.
+    pub gpus_per_node: Option<u32>,
+    /// Identifiers of QPUs requested for the job.
+    pub qpus: Option<Vec<String>>,
```
johnchildren, line 60: I don't expect we are going to be using this feature in the near future, but is it intended to be compatible with QRMI slurm resources?

tierkreis/src/orchestrator.rs
```diff
@@ -1235,6 +1240,7 @@ fn build_task_action(
             inputs,
             outputs,
             task_handle,
+            resources: HashMap::new(),
```
johnchildren, line 1244: I guess this is what we will populate later once we have resource configuration as an option/orchestration step?

tierkreis/src/runtime.rs
```diff
@@ -458,6 +468,7 @@ async fn executor_registry_from_config(
     config: &RuntimeConfig,
 ) -> Result<ExecutorRegistry, miette::Error> {
     let mut executor_registry: HashMap<String, Box<dyn Executor>> = HashMap::new();
+    let templates = ScriptTemplates::default();
```
johnchildren, line 471: I still find it slightly surprising that the templates are shared like this, are we expecting to use the same templates between different "Wrappers" ?

tierkreis/src/runtime.rs
```diff
@@ -135,6 +137,19 @@ enum ExecutorConfig {
         client_config: NexusClientConfig,
         output_storage_name: String,
     },
+    Hpc {
+        hpc_storage_name: String,
+        output_storage_name: String,
+        scheduler: HpcSchedulerConfig,
+        poll_interval_secs: Option<u64>,
```
johnchildren, line 144: Are we always expecting to be polling based for HPC? I could imagine a future system that is notification based.

tierkreis/src/executor/hpc/slurm.rs
```diff
@@ -0,0 +1,196 @@
...
+        job_ids: Vec<String>,
+    ) -> futures::future::BoxFuture<'_, Result<HashMap<String, SchedulerStatus>>> {
+        let scheduler = self.clone();
+        async move {
+            if job_ids.is_empty() {
+                return Ok(HashMap::new());
+            }
+            let job_ids = job_ids.join(",");
+            let output = Command::new(&scheduler.sacct)
+                .args([
```
johnchildren: It might be useful to either use the long form version of these flags or add a comment describing what they do, it's a bit tricky to guess what they are other than `-o` and `-j`

johnchildren (commented):
Some initial comments, I still need to look at the actual executor in detail.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+        .with_worker_command("mpiexec --allow-run-as-root uv run /mpi_worker/main.py");
+        // TODO: enable mpi environment
+
+        let stream = executor.listen()?;
+        executor.execute(task_plans).await?;
+
+        let events = stream.take(3).collect::<Vec<_>>().await;
+        dbg!(&events);
+        assert_eq!(events.len(), 3);
+        assert!(matches!(
```
johnchildren: You can use `assert_matches` now

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    use super::*;
+    use crate::{
+        asset_storage::{FileAssetStorage, assert_registry_contains_values, test_storage_registry},
+        event::{NodeEvent, NodeStatus, WorkflowRunEvent},
+        executor::{HPCExecutor, SlurmWrapper},
+    };
+    // Test that we can launch a task and listen for
+    // errors when they occur
+    #[tokio::test]
+    #[ignore = "Requires a local SLURM setup"]
```
johnchildren: Plausibly we should make this an integration test? https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    function_name: String,
+    inputs: HashMap<String, PathBuf>,
+    outputs: HashMap<String, PathBuf>,
+    output_dir: PathBuf,
+    done_path: PathBuf,
+    error_path: PathBuf,
+    logs_path: Option<PathBuf>,
+}
+
+struct EnvironmentCarrier<'a>(&'a mut HashMap<String, String>);
```
johnchildren: I think this exists in two places now, possibly worth putting into a `common` module for executors?

tierkreis/src/executor/hpc/slurm.j2
philipp-seitz: I added it for psij I guess  there's currently no point in having it anymore

tierkreis/src/executor/hpc/spec.rs
philipp-seitz: this is what is handed to --output, usually stdout

tierkreis/src/executor/hpc/spec.rs
philipp-seitz: this is what is handed to --error, usually stderr

tierkreis/src/runtime.rs
philipp-seitz: I think such an executor would work quite differently and would warrant a new variant

tierkreis/src/executor/hpc/spec.rs
philipp-seitz: yes

tierkreis/src/executor/hpc/spec.rs
johnchildren: I think the docstring could be clearer that it's the output of the job rather than the output of the scheduler.

tierkreis/src/executor/hpc/slurm.rs
```diff
@@ -0,0 +1,196 @@
...
+            templates: ScriptTemplates::default(),
+        }
+    }
+}
+
+impl SlurmWrapper {
+    /// Construct a wrapper for the repository's local Slurm Docker shims.
+    #[must_use]
+    #[cfg(test)]
+    pub fn local() -> Self {
```
johnchildren, line 68: Theoretically you might not need this if the binaries are in $PATH right? I would have assumed that when executing in the container this would be the case?

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+
+impl opentelemetry::propagation::Injector for EnvironmentCarrier<'_> {
+    fn set(&mut self, key: &str, value: String) {
+        self.0.insert(key.to_uppercase().replace('-', "_"), value);
+    }
+}
+
+struct BackgroundTaskPlan {
```
johnchildren, line 49: I guess there's a lot of overlap with the subprocess executor here as well.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+                }
+            }
+            else => break,
+        }
+    }
+}
+
+/// Event-based executor for subprocess-compatible workers.
+pub struct HPCExecutor {
+    scheduler: Arc<dyn SchedulerWrapper>,
```
johnchildren: I suspect you could make this a generic trait rather than using dynamic dispatch here as there should only be one SchedulerWrapper per HPCExecutor.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+        if !storage.contains_key(hpc_storage_name) {
+            return Err(miette!("hpc_storage_name not in registry"));
+        }
+        if !storage.contains_key(output_storage_name) {
+            return Err(miette!("output_storage_name not in registry"));
+        }
+        let tierkreis_dir = home_dir()
+            .unwrap_or_else(|| "/tmp".into())
+            .join(".tierkreis/tmp");
+        std::fs::create_dir_all(&tierkreis_dir)
```
johnchildren, line 386: I would probably use `TempDir` here instead (See the subprocess executor).

We do have an issue in the rust runtime though where we don't ever create a lot of directories in ~/.tierkreis/ that we are expecting to exist, so it would be good to define what directories we thing should exist and create them when the process starts if needed.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+            Some(
+                SchedulerStatus::Queued | SchedulerStatus::Complete | SchedulerStatus::Running,
+            ) => Ok(job_id),
+            _ => Err(miette!("Job not active")),
+        }
+    }
+}
+
+impl Executor for HPCExecutor {
+    // TODO: How to make sure this is run on the compute node?
```
johnchildren, line 535: I guess we could put something in the templates directory that we can spawn as a job? I guess it would need similar resources to a real job but with smaller requests (not sure how viable that is).

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    };
+    // Test that we can launch a task and listen for
+    // errors when they occur
+    #[tokio::test]
+    #[ignore = "Requires a local SLURM setup"]
+    async fn execute_hpc() -> miette::Result<()> {
+        let checkpoints_path = std::env::var_os("HOME")
+            .map(PathBuf::from)
+            .ok_or_else(|| miette!("HOME is not set"))?
+            .join(".tierkreis/slrm");
```
johnchildren: What's the deal with the new path here?

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    use crate::{
+        asset_storage::{FileAssetStorage, assert_registry_contains_values, test_storage_registry},
+        event::{NodeEvent, NodeStatus, WorkflowRunEvent},
+        executor::{HPCExecutor, SlurmWrapper},
+    };
+    // Test that we can launch a task and listen for
+    // errors when they occur
+    #[tokio::test]
+    #[ignore = "Requires a local SLURM setup"]
+    async fn execute_hpc() -> miette::Result<()> {
```
johnchildren, line 664: I think I don't really understand how this test is meant to be run, is it using `cargo test` when mounted inside the docker container?

tierkreis/src/executor/hpc.rs
johnchildren: Ah possibly this is what you were talking about with TempDir not working? I see this being used later with the `reserve_tmp_paths` method.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+        };
+        let context = tracing::Span::current().context();
+        opentelemetry::global::get_text_map_propagator(|propagator| {
+            propagator.inject_context(&context, &mut EnvironmentCarrier(&mut job_spec.environment));
+        });
+        let result = self.scheduler.submit(job_spec, script_path).await?;
+        Ok(result)
+    }
+
+    async fn is_job_active(&self, job_id: String) -> miette::Result<String> {
```
johnchildren: `active` kind of makes me think of `queued` or `running` but not `complete`. I guess the implication of this function is finding jobs that exist and haven't error'd? Not sure if I have a better work.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+        match self
+            .scheduler
+            .check(vec![job_id.clone()])
+            .await?
+            .get(&job_id)
+        {
+            Some(
+                SchedulerStatus::Queued | SchedulerStatus::Complete | SchedulerStatus::Running,
+            ) => Ok(job_id),
+            _ => Err(miette!("Job not active")),
```
johnchildren, line 529: Could be useful to distinguish between jobs that have error'd and jobs that don't exist.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+        task: &TaskPlan,
+        worker_args_path: &Path,
+        script_path: &Path,
+    ) -> miette::Result<String> {
+        let worker_command = self
+            .worker_command
+            .clone()
+            .unwrap_or_else(|| format!("tkr-{}", task.worker_name.replace('_', "-")));
+        let command = format!("{worker_command} {}", worker_args_path.display());
+        let hpc_resources = serde_json::from_value(task.resources.clone().into_iter().collect())
```
johnchildren, line 495: I would be surprised if you need all these `.clone().into_iter().collect()` calls?

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+            let task = tokio::task::spawn_blocking(|| {
+                let re = Regex::new(r"tkr-.*-worker")
+                    .into_diagnostic()
+                    .wrap_err("Failed to compile Worker name regex")?;
+                let paths = which_re(&re)
+                    .into_diagnostic()
+                    .wrap_err("Failed to search for Worker binaries")?;
+                Ok(paths
+                    .map(|path| WorkerSpec {
+                        worker_name: path.file_name().unwrap().to_str().unwrap().to_string(),
```
johnchildren: I guess this might be the case in the subprocess executors, but we probably should handle these errors instead of unwrapping.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+                    .tempfile_in(&self.tkr_tmp_dir)
+                    .into_diagnostic()?
+                    .keep()
+                    .into_diagnostic()
+                    .map(|(_, path)| path)
+            })
+            .collect()
+    }
+
+    async fn build_outputs(&self, outputs: HashSet<String>) -> miette::Result<OutputSpecs> {
```
johnchildren, line 444:
```suggestion
    async fn build_outputs(&self, outputs: &HashSet<String>) -> miette::Result<OutputSpecs> {
```

johnchildren: This should let you avoid the clone call in `execute`

tierkreis/src/executor/hpc.rs
johnchildren:
```suggestion
    async fn is_job_active(&self, job_id: &str) -> miette::Result<String> {
```

You end up cloning job_id in the body of the function anyway so you may as well just take a reference and then call `.to_string()` on it in the body (though I think that clone can probably be avoided as well).

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+                .try_lock()
+                .map_err(|err| miette!("Failed to listen: {}", err))?;
+
+            receiver.take().ok_or_else(|| {
+                miette!("Failed to listen: HPCExecutor is already being listened to.")
+            })?
+        };
+        Ok(channel.boxed())
+    }
+    fn cancel(
```
johnchildren, line 625:
```suggestion
    
    fn cancel(
```

Super nitpick but please include a line of whitespace between function definitions. (not sure if we can configure rustfmt to do this automatically).

johnchildren: https://github.com/rust-lang/rustfmt/issues/3382 seems like we would need to use nightly rustfmt for it...

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    async fn build_outputs(&self, outputs: HashSet<String>) -> miette::Result<OutputSpecs> {
+        let output_specs = reserve_asset_specs(
+            &self.asset_storage_registry,
+            &self.hpc_storage_name,
+            outputs.len(),
+        )
+        .await?;
+        let outputs: HashMap<_, _> = outputs.into_iter().zip(output_specs).collect();
+        let output_paths = outputs
+            .iter()
```
johnchildren, line 453: If you call `.into_iter()` here you can probably avoid the `name.clone()` in the `.map` call below.

tierkreis/src/executor/hpc.rs
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    fn reserve_tmp_paths(&self, count: usize) -> miette::Result<Vec<PathBuf>> {
+        (0..count)
+            .map(|_| {
+                tempfile::Builder::new()
+                    .prefix("tkr-")
+                    .tempfile_in(&self.tkr_tmp_dir)
+                    .into_diagnostic()?
+                    .keep()
```
johnchildren, line 437: I guess this an interesting point about temporary worker call args file in this case, we do want them to stay around if the runtime process ends (perhaps unlike the subprocess executor). Maybe this needs more thought when we look at the worker protocol again.

tierkreis/src/executor/hpc.rs
philipp-seitz: i used the same name as the nexus one 😬

tierkreis/src/executor/hpc.rs
philipp-seitz: the problem is that we can not know ahead of time in the case we have different cluster partitions with different configurations. we can only make that query with user input

tierkreis/src/executor/hpc.rs
philipp-seitz: this currently clashes with our python directory layout? also this would not change the fact that we need to ignore it for CI?

tierkreis/src/executor/hpc.rs
philipp-seitz: I wanted to mimic an hpc environment e.g. where this would be scratch. It's not strictly necessary

tierkreis/src/executor/hpc/slurm.rs
philipp-seitz: if we execute inside the container $PATH should resolve this correctly

tierkreis/src/runtime.rs
philipp-seitz: Originally I did it to be able to load the templates during runtime construction time

johnchildren (approved):

PR #638, opened 2026-09-08
chore: Remove N+1 queries pattern in node queries

johnchildren:
Removes the number of queries for node outputs by batching up locations that have completed times and doing a single query with a hardcoded chunk size.

Code generated by gpt-6-asta.

### DIFF
--- a/tierkreis/src/state/queries.rs
+++ b/tierkreis/src/state/queries.rs
@@ -1,7 +1,7 @@
 /*!
 This module defines the queries for reading the workflow state from the `SQlite` database.
 */
-use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


tierkreis/src/state/queries.rs
```diff
@@ -425,6 +425,9 @@ pub async fn read_node_state(
     }
 }
 
+// Leave room for run/attempt parameters below SQLite's bind limit.
+const READ_BATCH_SIZE: usize = 500;
```
johnchildren, line 429: I'm not super convinced by the name for this constant if it is dependent on the run/attempt parameters and the bind limit.

PR #652, opened 2026-09-15
docs: Update documentation from new interface

johnchildren:


docs/source/examples/errors_and_debugging.ipynb
```diff
@@ -101,34 +86,18 @@
...
-    "storage.clean_graph_files()\n",
-    "with contextlib.suppress(TierkreisError):\n",
-    "    run_graph(\n",
-    "        storage,\n",
-    "        executor,\n",
-    "        error_graph(),\n",
-    "        {\"value\": \"world!\"},\n",
-    "        polling_interval_seconds=0.1,\n",
-    "    )"
+    "states = await runtime.debug_read_node_states(run_id, 0, [\"N0\"])\n",
```
johnchildren, line 89: I think this doesn't really capture the intention of the original docs, perhaps a closer example would be to get the error'd locations and then look at their node states?

docs/source/examples/first_graph.ipynb
```diff
@@ -294,8 +283,7 @@
     }
    },
    "source": [
-    "If we have already run this example then there will already be files at this directory in the storage.\n",
-    "If we want to reuse the directory then run"
+    "A newly-created in-memory runtime starts with no saved workflows or runs, so it does not require a workflow UUID or cleanup step.\n"
```
johnchildren, line 286: I think this is a mistake, new_default() will save checkpoints to the user's storage. We don't need to clean the files beforehand anymore however.

PR #661, opened 2026-09-16
experiment: Vibe coded rust cli

johnchildren:


tierkreis/Cargo.toml
```diff
@@ -10,10 +10,6 @@ documentation = "https://quantinuum.github.io/tierkreis/"
 [lib]
 crate-type = ["rlib", "cdylib"]
 
-# Disabled until feature-parity with the existing CLI tool.
```
johnchildren, line 13: I would rather it had enabled this rather than deleting the block

johnchildren:
This does feel a lot like "plausible output" but from manual testing it seems quite broken.

PR #662, opened 2026-09-17
chore: Split Orchestrator into multiple components

johnchildren:
Some of the doc strings look wrong but could be a decent starting point.


johnchildren:
Some of these changes look reasonable, a few docstrings are incorrect but I think this could be merge-able with a few human changes. I think moving the tests to the runtime module instead of having a compatibility layer for the orchestrator tests would be one obvious improvement.

PR #663, opened 2026-09-17
chore: Refactor and cleanup action plans

johnchildren:


johnchildren:
I think the ActionPlan changes are quite good, some of the assignment to executors isn't quite what I had in mind but that can probably be left out.

